Issue #40, opened 2014-08-04
Refactor RedisPool to deal with RedisConnections and not Redis instances


popravich:
`RedisPool` must be pool of connections and `Redis` must hide from end-user what it uses inside.
It must be possible to instantiate `Redis` with connection and with connection's pool
thus providing single consistent high-level interface.

popravich:
Done in #129

PR #59, opened 2015-04-08
When an aioredis coroutine is wrapped in a cancelled task, the pending call is not cancelled

Natim:
When a coroutine is wrapped in a Task that is cancelled, the pending redis call is not cancelled and its result is lost.

Here is an example using ``BLPOP``

``` python
#!/usr/bin/env python

import asyncio
import aioredis


@asyncio.coroutine
def simulate_client():
    redis = yield from aioredis.create_redis(('localhost', 6379))
    print("$ Client sleeps")
    yield from asyncio.sleep(2)
    print("$ Client add task hello")
    yield from redis.lpush("channel", "hello")
    print("$ Client sleeps")    
    yield from asyncio.sleep(10)


@asyncio.coroutine
def simulate_server():
    redis = yield from aioredis.create_redis(('localhost', 6379))
    print("# Wait for task")
    try:
        data = yield from asyncio.wait_for(redis.blpop("channel"), 1)
    except asyncio.TimeoutError:
        print("# BLPOP timed-out")
        yield from asyncio.sleep(5)
        print("# Wait for task")
        try:
            data = yield from asyncio.wait_for(redis.blpop("channel"), 1)
        except asyncio.TimeoutError:
            print("# BLPOP timed-out the task is lost.")
        else:
            print("# BLPOP returned: %s" % data)
    else:
        print("# BLPOP returned: %s" % data)


asyncio.get_event_loop().run_until_complete(
    asyncio.wait({simulate_server(), simulate_client()})
)
```

For this example, I can fix it using the timeout value of BLPOP but I have the same behavior when using it with ``asyncio.wait``

```  python
    yield from asyncio.wait(
                {redis.blpop("channel"), websocket.recv()},
                return_when=asyncio.FIRST_COMPLETED)
```

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -246,6 +246,7 @@ class RedisFuture(asyncio.Future):
 
     def cancel(self):
         super(RedisFuture, self).cancel()
-        if self._conn._closed:
-            return
-        self._conn._reader_task.cancel()
+        if self._conn._reader_task:
+            self._conn._reader_task.cancel()
+            self._conn._reader_task = asyncio.Task(self._conn._read_data(),
+                                                   loop=self._conn._loop)
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -151,7 +151,7 @@ class RedisConnection:
             cb = None
         if encoding is _NOTSET:
             encoding = self._encoding
-        fut = asyncio.Future(loop=self._loop)
+        fut = RedisFuture(self, loop=self._loop)
         self._waiters.append((fut, encoding, cb))
         self._writer.write(encode_command(command, *args))
         return fut
@@ -237,3 +237,15 @@ class RedisConnection:
     @asyncio.coroutine
     def get_atomic_connection(self):
         return self
+
+
+class RedisFuture(asyncio.Future):
+    def __init__(self, conn, *args, **kwargs):
+        super(RedisFuture, self).__init__(*args, **kwargs)
+        self._conn = conn
+
+    def cancel(self):
+        super(RedisFuture, self).cancel()
+        if self._conn._closed:
+            return
+        self._conn._reader_task.cancel()
--- a/tests/_testutil.py
+++ b/tests/_testutil.py
@@ -25,9 +25,13 @@ def run_until_complete(fun):
     @wraps(fun)
     def wrapper(test, *args, **kw):
         loop = test.loop
-        ret = loop.run_until_complete(
-            asyncio.wait_for(fun(test, *args, **kw), 15, loop=loop))
-        return ret
+        try:
+            ret = loop.run_until_complete(
+                asyncio.wait_for(fun(test, *args, **kw), 15, loop=loop))
+        except asyncio.CancelledError:
+            pass
+        else:
+            return ret
     return wrapper
 
 


Natim:
Here is my fix proposal.

aioredis/connection.py
```diff
@@ -237,3 +237,16 @@ def auth(self, password):
...
+class RedisFuture(asyncio.Future):
+    def __init__(self, conn, *args, **kwargs):
+        super(RedisFuture, self).__init__(*args, **kwargs)
+        self._conn = conn
+
+    def cancel(self):
+        super(RedisFuture, self).cancel()
+        if self._conn._reader_task:
+            self._conn._reader_task.cancel()
+            self._conn._reader_task = asyncio.Task(self._conn._read_data(),
```
popravich: this is bad, because cancelling `_reader_task` will also cancel `_reader.read()` inside it and make connection unusable. So re-createing new `_reader_task` does not make much sense.

Natim: It is strange because test are passing. Could you add a test to show this behavior?

tests/_testutil.py
```diff
@@ -25,9 +25,13 @@ def run_until_complete(fun):
     @wraps(fun)
     def wrapper(test, *args, **kw):
         loop = test.loop
-        ret = loop.run_until_complete(
-            asyncio.wait_for(fun(test, *args, **kw), 15, loop=loop))
-        return ret
+        try:
+            ret = loop.run_until_complete(
+                asyncio.wait_for(fun(test, *args, **kw), 15, loop=loop))
+        except asyncio.CancelledError:
```
popravich: you should not catch `CancelledError` here -- the test must either run as expected or fail with error

Natim: Yes I had a doubt about this...

Natim:
Ok so continuing to think this through, I have come to the following piece of code:

``` python
#!/usr/bin/env python
import asyncio
import hiredis
from aioredis.util import encode_command
from aioredis.errors import ProtocolError, ReplyError


@asyncio.coroutine
def run_me():
    reader, writer = yield from asyncio.open_connection("localhost", 6379)
    writer.write(encode_command("BLPOP", "channel", 0))
    try:
        task = asyncio.Task(reader.read(65536))
        data = yield from asyncio.wait_for(task, 1)
    except asyncio.TimeoutError:
        task.cancel()
        data = yield from reader.read(65536)
    finally:
        parser = hiredis.Reader(protocolError=ProtocolError,
                                replyError=ReplyError)
        parser.feed(data)
        return parser.gets()

data = asyncio.get_event_loop().run_until_complete(run_me())
print(data)
```

The question to answer is how can we stop the current ``read()`` call in order to read the reader again later.

With this code, the python answer is:

```
RuntimeError: read() called while another coroutine is already waiting for incoming data
```

popravich:
With single connection this is very hard.
I'm trying to rewrite `Redis` #40  to make it work with pool of connection.
In that case such connection could be safely dropped and created new one.

I'd recommend you try using pool and not single connection.

Natim:
i think it is a bug in the StreamReaded actually that doesn't handle the cancel() correctly.

popravich:
No, the StreamReader is ok.
The problem is that you restart it in `except` block -- waiter future is cancelled but code in `finally` block has not been executed yet and you trying to restart reader, see [StreamReader._wait_for_data](https://hg.python.org/cpython/file/283036fe48e2/Lib/asyncio/streams.py#l382)

popravich:
Connection is blocked by Redis server until BLPOP timeout is expired. In case timeout is set to 0 the connection will be blocked forever and so simply must be closed.

Natim:
Yes ok it makes sense. So the right solution is to close the connection in case of a read for subscribe or BLPOP.

Maybe we should make sure that aioredis doesn't queue any commands for this long polling commands?

popravich:
Yes, and that should be done in terms of #40 -- pool should be hidden inside Redis and so it would be possible to choose free connection for new commands

Natim:
Right perfect, thanks a lot @popravich for tackling this, I am here to help in case you need me.

PR #62, opened 2015-04-30
zrevrangebyscore command fix

kernel0:
parameter order was wrong. Must be 'ZREVRANGEBYSCORE key max min'
ref : http://redis.io/commands/zrevrangebyscore


popravich:
Thanks

PR #84, opened 2015-10-09
Fix #83: wipe closed connection from free pool

### LINKED ISSUE #83: Drop closed connections on Pool.acquire
Currently on redis server restart the poll is full of closed connections.

That causes assertion error on acquiring (https://github.com/aio-libs/aioredis/blob/master/aioredis/pool.py#L117)

I suggest dropping closed connections in `.fill_free` coroutine.

asvetlov:


popravich:
Thanks

PR #85, opened 2015-10-09
Pep492

asvetlov:
1. Fix regression, `with await pool as conn` now works under Python 3.5
2. Add explicit support for async context manager `async with pool.get() as conn`. I doubt is `.get()` the best name but it can be changed easy.


popravich:
looks good, thanks)

PR #86, opened 2015-10-10
Implement iscan, izscan, isscan, ihscan commands

asvetlov:
Support `async for` in scan commands

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,7 @@ before_install:
 
 install:
 - pip install hiredis
-- pip install pyflakes
-- pip install pep8
+- pip install flake8
 - pip install coverage
 - pip install coveralls
 - pip install -e .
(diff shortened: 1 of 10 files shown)


asvetlov:
@popravich please review

popravich:
It looks like `_*ScanIter` classes are almost the same, I think it would be better to leave just few of them and place in `util.py`

asvetlov:
@popravich fixed

popravich:
nice)

asvetlov:
Cool! I have nothing more on my plate.
Could you make a release?

popravich:
WIll do release in a day or two, still need to do #87

asvetlov:
Sounds good!

popravich:
I just had a little time) so it published

PR #88, opened 2015-10-19
Close all pubsub channels when connection is closed

ciscorn:
Without this, Channel#wait_message() doesn't return nor raise any exceptions when the connection is lost.

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -267,7 +267,8 @@ class RedisConnection:
                 waiter.cancel()
             else:
                 waiter.set_exception(exc)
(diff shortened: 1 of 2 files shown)


popravich:
thanks

asvetlov:
@popravich please update CHANGES.txt also

popravich:
I'm doing this before each release

asvetlov:
Well, that's up to you. I prefer update CHANGES after every merge to don't forget something important.

popravich:
True, but don't worry, I do `git log <old_tag>..HEAD` every time to check everything)

PR #89, opened 2015-10-21
Async for pubsub Channel

popravich:
Added `Channel.iget()` method.

(travis caching updated as a side effect)


asvetlov:
I feel something wrong with `iget` name because the method does not the same as `get`.
Maybe `iter`?
Ideal solution would be adding `__aiter__` to `Channel` but we need to pass `encoding` and `decoder` unfortunately.

popravich:
True, it bothers me the same :/
`iter` is better but its the name of built-in function.
Anyway, if no other options appear I will rename it to `iter` later today

PR #107, opened 2016-02-09
Sentinel support

jettify:
Few months ago @bdrosen96 created sentinel support for aioredis. I have added minor code tweaks and just created this PR. This code we use successfully in production. 

There are fail over tests cases, I need your advise on how automate them in context of travisci.

### DIFF
--- a/aioredis/commands/__init__.py
+++ b/aioredis/commands/__init__.py
@@ -14,6 +14,7 @@ from .scripting import ScriptingCommandsMixin
 from .server import ServerCommandsMixin
 from .pubsub import PubSubCommandsMixin
 from .cluster import ClusterCommandsMixin
+from .sentinel import SentinelCommandsMixin
 
 __all__ = ['create_redis', 'Redis', 'Pipeline', 'MultiExec']
 
@@ -52,7 +53,8 @@ class Redis(GenericCommandsMixin, StringCommandsMixin,
             HashCommandsMixin, TransactionsCommandsMixin,
             SortedSetCommandsMixin, ListCommandsMixin,
             ScriptingCommandsMixin, ServerCommandsMixin,
-            PubSubCommandsMixin, ClusterCommandsMixin):
+            PubSubCommandsMixin, ClusterCommandsMixin,
+            SentinelCommandsMixin):
     """High-level Redis interface.
 
     Gathers in one place Redis commands implemented in mixins.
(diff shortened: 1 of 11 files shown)


jettify:
One possible way to make automatic failover tests is to leverage docker features on travis ci and then use something like docker-py to kill container from test.

popravich:
Thanks for PR, guys)

aioredis/commands/__init__.py
```diff
@@ -52,7 +53,8 @@ class Redis(GenericCommandsMixin, StringCommandsMixin,
             HashCommandsMixin, TransactionsCommandsMixin,
             SortedSetCommandsMixin, ListCommandsMixin,
             ScriptingCommandsMixin, ServerCommandsMixin,
-            PubSubCommandsMixin, ClusterCommandsMixin):
+            PubSubCommandsMixin, ClusterCommandsMixin,
+            SentinelCommandsMixin):
```
popravich: I don't think that `SentinelCommandsMixin` should be, well, mixed with other commands as Redis Sentinel is not a Redis. It support only few redis commands.
So I think it would be better to make separate `RedisSentinel` class and combine needed mixins

aioredis/commands/sentinel.py
```diff
@@ -0,0 +1,140 @@
+import asyncio
+
+from aioredis.util import wait_ok
+
+
+def nativestr(x):
```
popravich: We can get rid of this by passing `encoding` parameter to `execute` method and fixing `util.decode` function to recursively decode lists

popravich:
As for tests and travis I think we can start redis sentinel in .travis-ci.yml

popravich:
So, I created same branch (sentinel2) in aioredis repo and fixed tests to run on travis

Axik:
Any chances to get this in master?

popravich:
I think I'll be able to merge new pool with sentinel in a two or three weeks

Abdeljelil:
Hi, 
I tried a simple example to test sentinel with asyncio and I have this error when I try to call discover_master method. Is it a bug ?
 
File "/lib/python3.5/site-packages/aioredis/sentinel.py", line 251, in discover_master
    masters = yield from sentinel.sentinel_masters()
AttributeError: 'Redis' object has no attribute 'sentinel_masters'

This is my example:

```python
@asyncio.coroutine
def get_master_redis(loop, replication):

    redis_sentinel = yield from aioredis.create_sentinel(
        nodes, loop=loop
        )

    master = yield from redis_sentinel.discover_master(replication)

    return master
```

bdrosen96:
Redis object should include SentinelCommandsMixin mixin and that mixin does have def named sentinel_masters that takes only self.

popravich:
I'm closing it in favor of #181

Issue #108, opened 2016-02-28
RedisConnection.execute should raise custom exception when server is unreachable.


hotsyk:
Currently, RediscConnection.execute asserting live connection: https://github.com/aio-libs/aioredis/blob/master/aioredis/connection.py#L211 

As a result, if the connection to server lost (server restarted, or network lost), AssertionError will be raised. 
This leads to requirement to catch AssertionError if client would like to reconnect to server, and this is not correct way to catch this - more appropriate way would be to raise custom exception here and catch this specific exception in the client.

jettify:
Good point, also worth mention that asserts does not work if `python` run with `-O` or `-OO` flags

asvetlov:
@hotsyk would you provide a PR for replacement `AssertionError` with custom exception class inherited from `RedisError`?

A test is required, it may use mocks though. Redis server stopping is not needed, just simulate the case by mocking.

popravich:
It looks like `execute_pubsub` is missing even assertion, need to fix this too.

hotsyk:
Will do in a few days

popravich:
Great, thanks

popravich:
`execute_pubsub` fixed

PR #109, opened 2016-02-28
[#108] Raise ConnectionClosedError if connection lost instead of Asse…

hotsyk:
Raise ConnectionClosedError if connection lost instead of AssertionError


popravich:
Thanks

PR #116, opened 2016-03-24
Added support for SSL/TLS

moserware:
[By design](http://redis.io/topics/security), Redis doesn't natively support encryption via SSL/TLS. However, it can be added with an SSL proxy. 

For this reason, [some popular Redis libraries support SSL connections](https://github.com/andymccurdy/redis-py/blob/1c2071762ad9b9288e786665990083e61c1cf355/redis/connection.py#L656).

This change adds support for SSL/TLS by leveraging asyncio's existing SSL support in [asyncio.BaseEventLoop.create_connection](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.BaseEventLoop.create_connection) and using its semantics.

I didn't exhaustively research or test this change, so this pull request is simply my first pass at how such a change might look. I welcome critical feedback on it.

Thanks for this library!

### DIFF
--- a/aioredis/pool.py
+++ b/aioredis/pool.py
@@ -11,7 +11,7 @@ PY_35 = sys.version_info >= (3, 5)
 
 
 @asyncio.coroutine
-def create_pool(address, *, db=0, password=None, encoding=None,
+def create_pool(address, *, db=0, password=None, ssl=None, encoding=None,
                 minsize=10, maxsize=10, commands_factory=Redis, loop=None):
     """Creates Redis Pool.
 
@@ -24,7 +24,7 @@ def create_pool(address, *, db=0, password=None, encoding=None,
     Returns RedisPool instance.
     """
 
-    pool = RedisPool(address, db, password, encoding,
+    pool = RedisPool(address, db, password, ssl, encoding,
                      minsize=minsize, maxsize=maxsize,
                      commands_factory=commands_factory,
                      loop=loop)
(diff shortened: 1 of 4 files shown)


asvetlov:
Please fix pep8 errors and add tests

popravich:
Thanks for PR,
this looks good except few [flake8 complaints](https://travis-ci.org/aio-libs/aioredis/jobs/118243809#L265-L267).
I will take a closer look in a day or two.

moserware:
@asvetlov @popravich Thanks for the feedback! I fixed the PEP8 whitespace issues.

Regarding testing, I've only been using the `ssl=True` scenario (which uses defaults and works with a trusted certificate authority issued certificate)

It looks like when @oranagra of RedisLabs [added SSL support to redisy-py](https://github.com/andymccurdy/redis-py/pull/446#issuecomment-42936052), he just noted that it worked with an SSL server.

One option is to create a self-signed certificate and then use `stunnel` locally. Perhaps with this setup, one could manually create the [ssl.SSLContext](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) and trust the generated certificate.

Any guidance on where I should go from here? I didn't know if you'd want to add something like stunnel to your testing environment.

Thanks again for your consideration of this PR!

aioredis/pool.py
```diff
@@ -36,13 +36,14 @@ class RedisPool:
     """Redis connections pool.
     """
 
-    def __init__(self, address, db=0, password=None, encoding=None,
+    def __init__(self, address, db=0, password=None, ssl=None, encoding=None,
```
popravich: The `ssl` parameter here must be keyword-only argument.

moserware:
Thank you for cleaning this up and merging it! I look forward to using it via `pip`

popravich:
I will do release this week.

moserware:
:+1: That'd be great. Thanks again!

PR #117, opened 2016-03-28
Tests for ssl parameter

popravich:


moserware:
Thank you for cleaning this up and merging it! I look forward to using it via `pip`

PR #118, opened 2016-04-05
Bugfix: Shield close_waiter from cancellation of wait_closed.

the-panda:
Otherwise, if an 'await connection.wait_closed()' is cancelled, the _close_waiter will be cancelled, too. This raises an error, because when _read_task is done, it tries to set the result of the already cancelled _close_waiter.


popravich:
Please add tests

the-panda:
I added a basic test for wait_closed and one that fails without the fix above.

popravich:
Thanks

Issue #123, opened 2016-05-13
Redis queue class based on rpush/lpop


shicky:
For my application, I have redis queue class that wraps around rpush/lpop.
Is it possible to include in util.py?

ushuz:
`aioredis` is a general purpose Redis driver / connector, while queue is a specific purpose application. Queue is not what a driver needs to do its job, so I think it shouldn't be included in `aioredis`.

shicky:
Perhaps you are right.
But I still think it is common use-case for Redis and wouldn't hurt to be included.

popravich:
Hi guys,
I agree with ushuz that such `Queue` should not be included in `aioredis`,
because its hard to satisfy everybodys needs: someone may want Queue based on lists
someone may want it based on sorted sets, someone may want it to be built with some other concept.

Anyway, you can publish you Queue implementation on pypi and give me a link to be included in docs.
I think it would be really helpful for lib users to see a list with such extensions.

popravich:
As for PR, it has some inconsistencies and may not work as expected, if you want I can comment what is wrong.

shicky:
Thank you Alexey.
Could you please provide some feedback on current implementation?

Best Regards,

Steve (Sick) Yoon
(647) 760 - 2280

PR #125, opened 2016-05-13
redis queue implementation (LPOP, BLPOP, RPUSH, LLEN)

shicky:
## What these changes does?
* Add Queue class that wraps around LPOP, BLPOP, RPUSH, LLEN

## How to test your changes?
[https://github.com/shicky/aioredis/blob/redis_queue/tests/queue_test.py](https://github.com/shicky/aioredis/blob/redis_queue/tests/queue_test.py)

## Related issue number
#123 

## Checklist

- [ O ] Code is written and well
- [ O ] Tests for the changes are provided
- [ X ] Documentation reflects the changes

### DIFF
--- a/aioredis/util.py
+++ b/aioredis/util.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import sys
+import uuid
 
 from .errors import ChannelClosedError
 from .log import logger
@@ -8,7 +9,6 @@ from .log import logger
 PY_35 = sys.version_info >= (3, 5)
 
 _NOTSET = object()
-
 
 # NOTE: never put here anything else;
 #       just this basic types
@@ -18,7 +18,7 @@ _converters = {
     str: lambda val: val.encode('utf-8'),
     int: lambda val: str(val).encode('utf-8'),
     float: lambda val: str(val).encode('utf-8'),
-    }
+}
 
 
 def _bytes_len(sized):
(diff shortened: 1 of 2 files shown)


aioredis/util.py
```diff
@@ -55,6 +55,62 @@ def decode(obj, encoding):
...
+            yield from self._conn.rpush(self._key, data)
+            return True
+        return False
+
+    @asyncio.coroutine
+    def put_nowait(self, data):
+        if not (yield from self.full()):
+            asyncio.ensure_future(self._conn.rpush(self._key, data))
+            return True
+        return False
```
popravich: `put()` and `put_nowait()` methods are the main problem:
the main idea is ok -- "check if list is not full and then add one item",
but implementation would not work this way as between list size check and data append command
list may change in redis, consider this timeline:
```
coro1: -- llen() ---------- rpush(data) ----->
coro2: ---------- llen() --------------- rpush(data) --->
```
So the best way to implement constrained queue would be to use Lua scripting.
Redis executes lua-script as single command so you can check the len and add (or not add) data
to list atomically.

popravich: Having both `put` and `put_nowait` mean that first one must block until data is actually added to list (when it has space), however it wont happen, there should be a loop wating until list `is not full()`, then adding data to it and returning True.

aioredis/util.py
```diff
@@ -55,6 +55,62 @@ def decode(obj, encoding):
     return obj
 
 
+class Queue:
+
+    def __init__(self, connection, key=uuid.uuid4().hex, max_size=0, timeout=None):
```
popravich: Having such default value for key is **totally wrong**.
If someone using this Queue will not specify `key` parameter he will end up working with different queues every time application restarts.
key argument must not have default value.

aioredis/util.py
shicky: Thank you for the feedback.
Are you suggesting that it would be better to use lua script to implement this and wrap it with python using lupa?

popravich: No, I mean using redis lua scripting -- http://redis.io/commands/eval

PR #130, opened 2016-06-18
Extended HMSET

aamalev:
Implementation

### DIFF
--- a/aioredis/commands/hash.py
+++ b/aioredis/commands/hash.py
@@ -56,13 +56,20 @@ class HashCommandsMixin:
 
     def hmset(self, key, *args, **kwargs):
         """Set multiple hash fields to multiple values."""
-        if args:
+        if len(args) == 1 and isinstance(args[0], dict):
+            is_dict = True
+        elif args:
             is_dict = tuple(map(lambda x: isinstance(x, dict), args))
         elif not kwargs:
             raise TypeError("length of args or kwargs must be > 0")
 
         if not args:
             pairs = functools.reduce(operator.add, kwargs.items())
+
+        elif is_dict is True:
+            pairs = functools.reduce(operator.add, args[0].items())
+            if kwargs:
+                pairs = functools.reduce(operator.add, kwargs.items(), pairs)
 
         elif all(is_dict):
             acc = collections.OrderedDict()
--- a/aioredis/commands/hash.py
+++ b/aioredis/commands/hash.py
@@ -67,6 +67,8 @@ class HashCommandsMixin:
                     # key: '1' != b'1' != 1
                     acc = collections.OrderedDict()
                 acc.update(i)
+            elif acc is not None:
+                raise TypeError("not supported mixed type of args")
             else:
                 acc = None
                 pairs = args
--- a/aioredis/commands/hash.py
+++ b/aioredis/commands/hash.py
@@ -56,36 +56,30 @@ class HashCommandsMixin:
 
     def hmset(self, key, *args, **kwargs):
         """Set multiple hash fields to multiple values."""
-        if not args and not kwargs:
-            raise TypeError(
-                "length of args or kwargs must be > 0")
+        if args:
+            is_dict = tuple(map(lambda x: isinstance(x, dict), args))
+        elif not kwargs:
+            raise TypeError("length of args or kwargs must be > 0")
 
-        acc = None
-        for i in args:
-            if isinstance(i, dict):
-                if acc is None:
-                    # key: '1' != b'1' != 1
-                    acc = collections.OrderedDict()
+        if not args:
+            pairs = functools.reduce(operator.add, kwargs.items())
+
+        elif all(is_dict):
+            acc = collections.OrderedDict()
+            for i in args:
                 acc.update(i)
-            elif acc is not None:
-                raise TypeError("not supported mixed type of args")
-            else:
-                acc = None
-                pairs = args
-                break
-        else:
-            pairs = ()
+            if kwargs:
+                acc.update(kwargs)
+            pairs = functools.reduce(operator.add, acc.items())
 
-        if pairs and len(pairs) % 2 != 0:
+        elif any(is_dict):
+            raise TypeError("not supported mixed type of args")
+
+        elif len(args) % 2 != 0:
             raise TypeError("length of dicts_or_pairs must be even number")
 
-        if kwargs and not acc:
-            acc = kwargs
-        elif kwargs:
-            acc.update(kwargs)
-
-        if acc:
-            pairs = functools.reduce(operator.add, acc.items(), pairs)
+        else:
+            pairs = args
 
         return wait_ok(self._conn.execute(b'HMSET', key, *pairs))
 
--- a/aioredis/commands/hash.py
+++ b/aioredis/commands/hash.py
@@ -1,8 +1,11 @@
-import collections
-import functools
-import operator
-
-from aioredis.util import wait_ok, wait_convert, wait_make_dict, _NOTSET, PY_35
+from aioredis.util import (
+    wait_ok,
+    wait_convert,
+    wait_make_dict,
+    make_pairs,
+    _NOTSET,
+    PY_35,
+    )
 
 if PY_35:
     from aioredis.util import _ScanIterPairs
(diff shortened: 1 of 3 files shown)


popravich:
Hi, this is good feature but I dont like your implementation -- it is too complex and has bugs.
First of all, if you add this test it will fail:
```python
yield from redis.hmset('key', {'a': 1}, {'b': 2}, 'c', 3, d=4, a=5)
res = yield from redis.hgetall('key')
assert res == {b'a': b'5',
               b'b': b'2',
               b'c': b'3',
               b'd': b'4'}
```
Second, implementation, the public interface must be as follows:
` hmset(key, *args, **kwargs)`
internally this must be handled as follows:
* check `if not args and not kwargs` -> `TypeError`
* if `len(args) == 1` -> `args[0]` must be a `dict`
* or `len(args) % 2 == 0` -> args is a field-value pairs
* finally `if len(args) != 1 or len(args) % 2 != 0` -> `TypeError`
* add kwargs if any
* and execute `HMSET`

aamalev:
Hi @popravich. My implementation dicts_or_pairs not supported mixed types.

This test will fail in both cases in one and the same place.

    def execute(self, command, *args, encoding=_NOTSET):
             ...
    >       if None in set(args):
    E       TypeError: unhashable type: 'dict'

A similar bug in hset:
    yield from redis.hset('key', {'a': 1}, {'b': 2})

popravich:
> A similar bug in hset:
> yield from redis.hset('key', {'a': 1}, {'b': 2})

This is intended behaviour

> My implementation dicts_or_pairs not supported mixed types.

If so it should raise exception when mixing arg types

aamalev:
More successful implementation

aioredis/commands/hash.py
```diff
@@ -50,12 +54,34 @@ def hmget(self, key, field, *fields, encoding=_NOTSET):
...
-        return wait_ok(self._conn.execute(b'HMSET', key, field, value, *pairs))
+        if args:
+            is_dict = tuple(map(lambda x: isinstance(x, dict), args))
+        elif not kwargs:
+            raise TypeError("length of args or kwargs must be > 0")
+
+        if not args:
+            pairs = functools.reduce(operator.add, kwargs.items())
+
+        elif all(is_dict):
```
popravich: why do you need all `args` to be dicts?
what is the reason for calling hmset like this:
```python
redis.hmset(key, {'a': 1}, {'b': 2})
```
and not simply with one dict?

aioredis/commands/hash.py
```diff
@@ -50,12 +54,34 @@ def hmget(self, key, field, *fields, encoding=_NOTSET):
...
+            acc = collections.OrderedDict()
```
popravich: no need in OrderedDict, `args` are ordered, so every next pair in args will overwrite previous (with same key).

aamalev:
An example of the need OrderedDict

    redis.hmset(key, {'a': 1}, {b'a': 2}, a=3)

As the regular dict:
    {'a': 3, b'a':2} or {b'a': 2, 'a':3}

Another crazy example:
    redis.hmset(key, {1: 1}, {b'1': 2}, {'1': 3})

popravich:
OMG this is totaly crazy)
It is end-user's responsibility to control input data.
Ok, what is gonna happen in this case (which value will win)?
```python
redis.hmset(key, {b'a': 1, 'a': 2}, {'a': 3, b'a': 4})
```

kxepal:
I agree with @popravich: no need to be too smart. If conflicts happens, better to cry loudly instead of picking decision that user may not be aware/expected about. Only user may solve them right.

popravich:
I like it as simple as possible -- so it 100% clear how it works and what to expect.
Lets not break current `hmset` but create another one for dicts:
`hmset_dict(key, *args, **kwargs)` args allowed only to hold single a dict otherwise kwargs must be provided.
The implementation should simply unpack args[0] & kwargs into pairs and call original `hmset`

aamalev:
Convinced. Therefore, I propose a simple implementation.

aioredis/util.py
```diff
@@ -319,3 +321,26 @@ def create_future(loop):
...
+        pairs = make_flat(args[0].items())
+
+    elif len(args) % 2 != 0:
+        raise TypeError("length of args must be even number")
+
+    else:
+        pairs = args
+
+    if kwargs:
+        pairs = make_flat(kwargs.items(), pairs)
```
kxepal: What's a point of generating tuples for every iteration? Sounds like you need in:
```
pairs = list(pairs)
pairs.extend(chain.from_iterable(kwargs.items())
```

It's all said:
```
In [6]: %timeit make_flat(kwargs.items(), tuple())
100 loops, best of 3: 2.45 ms per loop

In [8]: %timeit list(chain.from_iterable(kwargs.items()))
10000 loops, best of 3: 78.2 µs per loop
```

aamalev: Cool, but what is your kwargs?
```
In [46]: kwargs = dict(enumerate(string.ascii_lowercase))

In [47]: %timeit make_flat(kwargs.items(), tuple())
100000 loops, best of 3: 11.1 µs per loop

In [48]: %timeit make_flat(kwargs.items())
100000 loops, best of 3: 10.3 µs per loop

In [49]: %timeit list(chain.from_iterable(kwargs.items()))
The slowest run took 5.11 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 6.89 µs per loop

In [50]: %timeit list(chain.from_iterable(kwargs.items()))
The slowest run took 5.22 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 6.93 µs per loop

In [51]: %timeit tuple(chain.from_iterable(kwargs.items()))
100000 loops, best of 3: 6.04 µs per loop```

Comment (kxepal) [inline aioredis/util.py]: `kwargs = {i:i for i in range(1000)}`

**Comment by popravich** [comment]
thanks

**Comment by popravich** [comment]
PR is squashed and merged.
However I've split `hmset` into original and `hmset_dict` to decrease ambiguity of arguments' combinations.

**Comment by aamalev** [comment]
hmset_dict not the correct name, as well as the method of supporting the pair. Share methods is that if someone thought to call the old way hmset: 

    redis.hmset(key, field='a', value='1', 'b', '2', 'c', '3')

But I do not think that someone so does

**Comment by aamalev** [comment]
make_pair can also be used in mset

**Comment by aamalev** [comment]
Sorry, of course kwargs at the end

    redis.hmset(key, 'b', '2', 'c', '3', field='a', value='1')

crazy?

**Comment by popravich** [comment]
> `redis.hmset(key, 'b', '2', 'c', '3', field='a', value='1')`

This does not work -- multiple values for positional args

**Comment by aamalev** [comment]
I try to challenge the motivation of separation methods.
This work on the old method. But I think not used someone this form

**Comment by popravich** [comment]
Ok, I want to keep original `hmset` as is because I don't want something that would fail in current version to success in newer.
I don't want old hmset to magically work if someone pass invalid args,
currently everyone using hmset must call it with field-value pairs positional arguments
and if someone try to call it with a dict -- will get TypeError.
Now imagine something like this:
```python
hopefully_its_pairs = get_something()
yield from redis.hmset(key, *hopefully_its_pairs)

# and assume `get_something` is a function with lots of if/else:
def get_something():
    if case_a:
        return 'field', 'value'
    elif case_b:
        return 'field2', 'value2'
    elif ...
    elif ...
    elif case_N:
        return {'field': "value"}
```
Such cases may not be covered with tests and/or may appear only in production environment.
So if hmset had changed to allowed dicts this bug would suddenly disappear.
And it sounds more like a joke -- fixing bug in user's code by upgrading 3rd-party library.

aamalev:
Reasonable.
I did not notice that hmset_dict only works with dict. For strictness propose to change the signature on 

    hmset_dict(key, fields: dict=None, **kwargs)

explicit arguments and simplified checks

popravich:
this may be conflicting with user's fields:
`hmset_dict(key, {'foo': 1, 'bar': 2}, fields='foo,bar')`

aamalev:
but current implementation already contains conflict

    yield from redis.hmset_dict(key, key='kw')

PR #131, opened 2016-07-04
fix util.decode to decode dicts also

alefteris:
Modified `aioredis.util.decode` to also decode dicts in responses recursively.

### DIFF
--- a/aioredis/util.py
+++ b/aioredis/util.py
@@ -54,6 +54,9 @@ def decode(obj, encoding):
         return obj.decode(encoding)
     elif isinstance(obj, list):
         return [decode(o, encoding) for o in obj]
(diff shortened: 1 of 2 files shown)


popravich:
I don't remember any commands returning dics. What are they?

alefteris:
@popravich `hgetall` returns a dict.

popravich:
Redis `HGETALL` command returns array reply,
aioredis `Redis.hgetall` method converts such array reply into dict, decoding is supported as well.

alefteris:
I'm so sorry. Somehow I didn't set the decode for Redis.hgetall and I thought there was an issue with the returned dict. Closing.

popravich:
no problem

PR #136, opened 2016-08-14
Close RedisPool when connection to the redis server has failed (To prevent asyncio lock)

cynecx:
This patch closes the RedisPool when the connection to the redis server has failed. We would get an asyncio error when closing the event loop:

```
Task was destroyed but it is pending!
task: <Task pending coro=<RedisPool._do_close() running at python3.5/site-packages/aioredis/pool.py:102> wait_for=<Future pending cb=[Task._wakeup()]>>
```

### DIFF
--- a/tests/pool_test.py
+++ b/tests/pool_test.py
@@ -391,8 +391,6 @@ def test_pool_close__used(create_pool, server, loop):
 
 @pytest.mark.run_loop
 def test_pool_check_closed_when_exception(create_pool, server, loop):
-    exception = False
-
     class TrackExceptionRecord(logging.Handler):
         def __init__(self):
             logging.Handler.__init__(self)
@@ -407,12 +405,8 @@ def test_pool_check_closed_when_exception(create_pool, server, loop):
     logger.setLevel(logging.DEBUG)
     logger.addHandler(handler)
 
-    try:
+    with pytest.raises(Exception):
         yield from create_pool(address=('127.3.2.1', 9999), loop=loop)
-    except:
-        exception = True
-
-    assert exception is True
 
     yield from asyncio.sleep(0, result=None, loop=loop)
     assert handler.has_closed_msg is True
--- a/tests/pool_test.py
+++ b/tests/pool_test.py
@@ -1,5 +1,7 @@
 import asyncio
+import logging
 import pytest
+import re
 
 from aioredis import (
     RedisPool,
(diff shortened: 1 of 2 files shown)


popravich:
Nice catch.
But, please provide a test:
try connecting to free port, catch error and check that pool is marked closed

cynecx:
@popravich Wait, I am not sure how to test this as `create_pool` raises an exception when the connection has failed, so it wouldn't return an instance of `RedisPool`. Hence there is no way to check if the pool is marked closed.

Another approach would be to catch exceptions and return the current state of pool.
What do you think?

popravich:
Yeah, missed that issue)
Lets do the following:
* add `logger.debug` call after `gather` in `_do_close()` with something like this
  `logger.debug("Closed %d connections", len(waiters))`
* in test try connecting to unused port
* catch exception
* run dummy loop iteration (so that `_do_closed()` finish)
* catch that log message
So we'll be sure that coroutine is done.

cynecx:
@popravich Isn't this a little bit inefficient to check? I mean, can't we simply check if the event loop tasks are marked done?.  Something like:

```
# ... create pool -> close

for task in asyncio.Task.all_tasks(loop):
    assert(task.done()) # All tasks should be "done" if the pool is successfully marked closed
```

popravich:
@cynecx this won't work as test itself is a task and is not done until test coroutine is ended.
With logging we'd be able to make even more complex test, for instance start redis server with `maxclients=1` and try to create pool with `min=2` and catch logs saying that 1 connection has been closed on that error

cynecx:
@popravich Can you review the latest commit. As a asyncio beginner I wasn't sure how to run a loop iteration while being in a loop, so I figured out using `asyncio.sleep` which seems hacky...
Also I wasn't sure how to catch debug messages, so I went with a logging handler.

tests/pool_test.py
```diff
@@ -385,3 +387,34 @@ def test_pool_close__used(create_pool, server, loop):
...
+            if re.match("Closed [0-9]+ connections", record.getMessage()):
+                self.has_closed_msg = True
+
+    handler = TrackExceptionRecord()
+    logger = logging.getLogger('aioredis')
+    logger.setLevel(logging.DEBUG)
+    logger.addHandler(handler)
+
+    try:
+        yield from create_pool(address=('127.3.2.1', 9999), loop=loop)
```
popravich:
```
with pytest.raises(Exception):
    yield from create_pool(...)
```

cynecx:
@popravich Please review the latest commit.

popravich:
Looks good. Thanks

PR #140, opened 2016-09-24
Set default and validate section for info cmd

alefteris:
Set a default param value of `'default'` (same as how redis-cli works http://redis.io/commands/INFO) for the `info()` command. Also added validation for the section param of the same command.

### DIFF
--- a/aioredis/commands/server.py
+++ b/aioredis/commands/server.py
@@ -12,6 +12,19 @@ class ServerCommandsMixin:
 
     SHUTDOWN_SAVE = 'SHUTDOWN_SAVE'
     SHUTDOWN_NOSAVE = 'SHUTDOWN_NOSAVE'
+    INFO_SECTIONS = [
+        'server',
+        'clients',
+        'memory',
+        'persistence',
+        'stats',
+        'replication',
+        'cpu',
+        'commandstats',
+        'cluster',
+        'keyspace',
+    ]
+    INFO_ARGS = INFO_SECTIONS + ['default', 'all']
 
     def bgrewriteaof(self):
         """Asynchronously rewrite the append-only file."""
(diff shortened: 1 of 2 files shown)


aioredis/commands/server.py
```diff
@@ -112,9 +125,21 @@ def flushdb(self):
...
+
+        :raises TypeError: if section is not string
+        :raises ValueError: if section is invalid
+
+        """
+        if not isinstance(section, str):
+            raise TypeError("section must be str")
+        section = section.lower()
+        if section not in self.INFO_ARGS:
+            raise ValueError("invalid section")
```
popravich: I'm ok with default and docstring update,
but I don't see any reasoning for `INFO_SECTIONS` and `INFO_ARGS`:
first is used only in test and the second constraints user's input.
The later is not realy good idea because if new sections appeared aioredis won't let users to query them.

popravich: Also checking `section` argument to be an instance of str is redundant -- it will be checked in `connection.execute` method.

alefteris: Good point about forward compatibility. I included the strict sections validation because I thought it would be more helpful to tell you exactly what was the issue when someone provides a wrong section.

I don't think there is any validation that the args should be a string done at `RedisConnection.execute()` and `util.encode_command()` just checks that it's one of bytes, str, int or float. Unless I missed something.

Right now, I just get some StopIteration exceptions if I provide an invalid section. Unless I'm doing something wrong.

Here is the example I used:

```python

async def go():
    redis = await aioredis.create_redis(('localhost', 32769), loop=loop)
    val = await redis.info(5)  # or redis.info('whatever')
    print(val)
    redis.close()

loop.run_until_complete(go())
```

And the error I get from the above:

```
  File "/Users/thanos/Dev/forks/aioredis/aioredis/util.py", line 205, in wait_convert
    return type_(result)
  File "/Users/thanos/Dev/forks/aioredis/aioredis/commands/server.py", line 229, in parse_info
    section = next(block)[2:].lower()
StopIteration

The above exception was the direct cause of the following exception:

  File "thanos.py", line 12, in <module>
    loop.run_until_complete(go())
  File "/Users/thanos/.pyenv/versions/3.5.2/lib/python3.5/asyncio/base_events.py", line 387, in run_until_complete
    return future.result()
  File "/Users/thanos/.pyenv/versions/3.5.2/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/Users/thanos/.pyenv/versions/3.5.2/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "thanos.py", line 8, in go
    val = await redis.info('')
RuntimeError: generator raised StopIteration
Task was destroyed but it is pending!
task: <Task pending coro=<RedisConnection._read_data() running at /Users/thanos/Dev/forks/aioredis/aioredis/connection.py:131> wait_for=<Future pending cb=[Task._wakeup()]> cb=[Future.set_result()]>
```

Or I could just update the PR and keep the default arg only if you think it's the best way to go.

Either way, thanks for your time to review it :)

popravich: OMG! `RuntimeError: generator raised StopIteration` that's totally incorrect behavior.
This is a bug in `parse_info` function and need to be fixed.

Anyway, I think checking `section` argument to be not False would be sufficient:
`if not section: raise ValueError(...)`
and no other constraints needed

popravich:
Thanks

PR #177, opened 2017-01-03
Geo commands

polyn0m:
Added GeoCommandsMixin for support Redis geo commands since redis 3.2.0

### DIFF
--- a/aioredis/commands/geo.py
+++ b/aioredis/commands/geo.py
@@ -1,0 +1,94 @@
+from aioredis.util import _NOTSET, wait_ok
+
+
+class GeoCommandsMixin:
+    """Geo commands mixin.
+
+    For commands details see: http://redis.io/commands#geo
+    """
+
+    def geoadd(self, key, longitude, latitude, member):
+        """Add one or more geospatial items in the geospatial index represented 
+        using a sorted set
+        """
+        return self._conn.execute(b'GEOADD', key, longitude, latitude, member)
+
+    def geodist(self, key, member1, member2, unit='m'):
+        """Returns the distance between two members of a geospatial index
+        """
+        return self._conn.execute(b'GEODIST', key, member1, member2, unit)
+
+    def geohash(self, key, member):
+        """Returns members of a geospatial index as standard geohash strings
+        """
+        return self._conn.execute(b'GEOHASH', key, member)
+
+    def geopos(self, key, member):
+        """Returns longitude and latitude of members of a geospatial index
+        """
+        return self._conn.execute(b'GEOPOS', key, member)
+
+    def georadius(self, key, longitude, latitude, radius, unit='m',
+                  with_coord=False, with_dist=False, with_hash=False,
+                  count=None, sort=None):
+        """Query a sorted set representing a geospatial index to fetch members 
+        matching a given maximum distance from a point
+
+        :raises TypeError: radius is not float or int
+        :raises TypeError: count is not float or int
+        :raises ValueError: if sort not equal ASC or DESC
+        """
+        args = []
+
+        if with_coord:
+            args.append('WITHCOORD')
+        if with_dist:
+            args.append('WITHDIST')
+        if with_hash:
+            args.append('WITHHASH')
+
+        if not isinstance(radius, (int, float)):
+            raise TypeError("radius argument must be int or float")
+        if count:
+            if not isinstance(count, int):
+                raise TypeError("count argument must be int")
+            args.append(count)
+        if sort:
+            if sort in ['ASC', 'DESC']:
+                raise TypeError("sort argument must be euqal ASC or DESC")
+            args.append(sort)
+
+        return self._conn.execute(b'GEORADIUS', key, longitude, latitude, radius, unit, *args)
+
+    def georadiusbymember(self, key, member, radius, unit='m',
+                          with_coord=False, with_dist=False, with_hash=False,
+                          count=None, sort=None):
+        """Query a sorted set representing a geospatial index to fetch members 
+        matching a given maximum distance from a member
+
+        :raises TypeError: radius is not float or int
+        :raises TypeError: count is not float or int
+        :raises ValueError: if sort not equal ASC or DESC
+        """
+        args = []
+
+        if with_coord:
+            args.append('WITHCOORD')
+        if with_dist:
+            args.append('WITHDIST')
+        if with_hash:
+            args.append('WITHHASH')
+
+
+        if not isinstance(radius, (int, float)):
+            raise TypeError("radius argument must be int or float")
+        if count:
+            if not isinstance(count, int):
+                raise TypeError("count argument must be int")
+            args.append(count)
+        if sort:
+            if sort in ['ASC', 'DESC']:
+                raise TypeError("sort argument must be euqal ASC or DESC")
+            args.append(sort)
+
+        return self._conn.execute(b'GEORADIUSBYMEMBER', key, member, radius, unit, *args)
--- a/aioredis/commands/geo.py
+++ b/aioredis/commands/geo.py
@@ -1,4 +1,4 @@
-from aioredis.util import _NOTSET, wait_ok
+from aioredis.util import wait_convert, wait_convert_with_opts, _NOTSET
 
 
 class GeoCommandsMixin:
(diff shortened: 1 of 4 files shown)


aioredis/commands/geo.py
```diff
@@ -0,0 +1,155 @@
...
+        return self._conn.execute(
+            b'GEOADD', key, longitude, latitude, member, *args, **kwargs
+        )
+
+    def geohash(self, key, member, *args, **kwargs):
+        """Returns members of a geospatial index as standard geohash strings
+        """
+        encoding = _NOTSET
+        if 'encoding' in kwargs:
+            encoding = kwargs.pop('encoding')
```
popravich: No need in such `encoding` parsing, it will be extracted from `kwargs` in `execute` method.

aioredis/commands/geo.py
```diff
@@ -0,0 +1,155 @@
...
+def pairs_float(value):
+    return [[float(val[0]), float(val[1])] for val in value]
+
+
+def geo_data_row(value, with_dist, with_coord, with_hash):
+    res_rows = []
+    for row in value:
+        res = []
+        
+        if with_dist and with_coord and with_hash:
```
popravich: This looks overcomplicated. Lets make it simple.

First of all I think it would be nice to have these results wrapped in `namedtuple`
so users would not have to remember meaning of each index in a row:
```python
GeoCoord = namedtuple('GeoCoord', 'longtitude latitude')
GeoRadius = namedtuple('GeoRadius', 'name distance hash coord')

def make_geo_radius(name, distance=None, hash=None, coord=None):
    if distance is not None:
        distance = float(distance)
    if hash is not None:
        hash = int(hash)
    if coord is not None:
        hash = GeoCoord(*map(float, coord))
    return GeoRadius(name, distance, hash, coord)
```
So `geo_data_row` can look like this:
```python
...
for row in value:
    if isinstance(row, list):
        res_rows.append(make_geo_radius(*row))
    else:
        res_rows.append(row)
...
```

aioredis/util.py
```diff
@@ -73,6 +73,14 @@ def wait_convert(fut, type_):
 
 
 @asyncio.coroutine
+def wait_convert_with_opts(fut, type_, **opts):
```
popravich: No need in separate function, just add `**kwargs` or `**opts` to `wait_convert`

aioredis/commands/geo.py
polyn0m: Removed

aioredis/util.py
polyn0m: Fixed

aioredis/commands/geo.py
polyn0m: Added types and rewrite function

aioredis/commands/geo.py
```diff
@@ -1,4 +1,10 @@
-from aioredis.util import wait_convert, wait_convert_with_opts, _NOTSET
+from collections import namedtuple
+
+from aioredis.util import wait_convert, _NOTSET
+
+
+GeoCoord = namedtuple('GeoCoord', ('longitude', 'latitude'))
+GeoMember = namedtuple('GeoRadius', ('member', 'dist', 'hash', 'coord'))
```
popravich: names should match

aioredis/commands/geo.py
```diff
@@ -112,44 +115,50 @@ def validate_georadius_options(radius, unit, count, sort_dir,
...
-        elif with_dist:
-            res.append(row[0])
-            res.append(float(row[1]))
-        elif with_hash:
-            res.append(row[0])
-            res.append(int(row[1]))
-        elif with_coord:
-            res.append(row[0])
-            res.append([float(row[1][0]), float(row[1][1])])
+        member, distance, coord, hash_ = None, None, None, None
```
popravich: Ok, my previous example was incorrect.
But I realy don't like this long if-elif noodles.
How about this:
```python
if isinstance(row, list):
    name = row.pop(0)
    dist = hash = coord = None
    if with_dist:
        dist = float(row.pop(0))
    if with_hash:
        hash = int(row.pop(0))
    if with_coord:
        coord = GeoCoord(*map(float, row.pop(0)))
    return GeoMember(name, dist, hash, coord)
```

tests/geo_commands_test.py
```diff
@@ -83,105 +87,143 @@ def test_geopos(redis):
...
-        'Palermo', 15.087269, 37.502669, 'Catania'
+        13.361389, 38.115556, 'Palermo', 15.087269, 37.502669, 'Catania'
     )
     assert res == 2
 
     res = yield from redis.georadius(
         'geodata', 15, 37, 200, 'km', encoding='utf-8'
     )
     assert res == [
-        ['Palermo'], ['Catania']
```
popravich: I think `georadius*` commands without `with_*` flags must behave same as Redis — return list of strings,
not list of lists of strings or namedtuples.

aioredis/commands/geo.py
polyn0m: @popravich what kind names must match?

tests/geo_commands_test.py
polyn0m: i think command must return same result - list of namedtuples, because no reason for return list of strings, it's same record without additional data.

aioredis/commands/geo.py
polyn0m: Ahhh, typos...

popravich:
build is still failing, please fix flake errors.

polyn0m:
Manual run result http://joxi.ru/aD2PGWpiNld9m3

tests/geo_commands_test.py
popravich: @polyn0m, I was thinking about it and still sceptical.
We must follow Redis `georadius` return convention, ie:
return list of strings when no `withcoord / withhash / withdist` flags are set;
return list of (named)tuples when any of above flags are set.

In cases when no `WITH*` flags are set distance, coordinate and hash fields
of tuple would be empty and the only usable field would be name.
User would need to convert such list of one-item-tuples to plain list
to use it (for instance to do join — `', '.join(['Palermo', 'Catania'])`,
or to pass this results to other DB query...)

In case we always return list of tuples there would be situations when double conversion
happen — aioredis convert plain list to list of tuples; then user converts list of tuples to plain list.
I think this would be uncommon case but it (returning plain list of strings) would
make code more readable and simplier (both library- and user- sides).

tests/geo_commands_test.py
```diff
@@ -0,0 +1,273 @@
...
+    res = yield from redis.geopos('geodata', 'Palermo')
+    assert res == [
+        GeoCoord(longitude=13.36138933897018433, latitude=38.11555639549629859)
+    ]
+
+    res = yield from redis.geopos('geodata', 'Catania', 'Palermo')
+    assert res == [
+        GeoCoord(longitude=15.087267458438873, latitude=37.50266842333162),
+        GeoCoord(longitude=13.36138933897018433, latitude=38.11555639549629859)
+    ]
```
popravich: Add test (and fix) for the following case:
```python
res = yield from redis.geopos('geodata', 'Catania', 'NonExistingMember', 'Palermo')
```

tests/geo_commands_test.py
```diff
@@ -0,0 +1,273 @@
...
+    assert res == [
+        GeoMember(
+            member='Palermo', dist=0.0, hash=3479099956230698,
+            coord=GeoCoord(13.361389338970184, 38.1155563954963)
+        ),
+        GeoMember(
+            member='Catania', dist=166.2742, hash=3479447370796909,
+            coord=GeoCoord(15.087267458438873, 37.50266842333162)
+        )
+    ]
```
popravich: Please add tests for each command without `encoding`

aioredis/commands/geo.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def geodist(self, key, member1, member2, unit='m'):
+        """Returns the distance between two members of a geospatial index
+        """
+        fut = self._conn.execute(b'GEODIST', key, member1, member2, unit)
+        return wait_convert(fut, float)
+
+    def georadius(self, key, longitude, latitude, radius, unit='m',
+                  dist=False, hash=False, coord=False,
+                  count=None, sort=None, encoding=_NOTSET):
```
popravich: Please keep `with_` prefixes for dist, hash and coord parameters.

aioredis/commands/geo.py
```diff
@@ -0,0 +1,143 @@
...
+        """Query a sorted set representing a geospatial index to fetch members
+        matching a given maximum distance from a point
+
+        :raises TypeError: radius is not float or int
+        :raises TypeError: count is not float or int
```
popravich: typo: `count` can be int only

aioredis/commands/geo.py
```diff
@@ -0,0 +1,143 @@
...
+            unit, *args, encoding=encoding
+        )
+        return wait_convert(
+            fut, make_geomember,
+            with_dist=dist, with_hash=hash, with_coord=coord
+        )
+
+    def georadiusbymember(self, key, member, radius, unit='m',
+                          dist=False, hash=False, coord=False,
+                          count=None, sort=None, encoding=_NOTSET):
```
popravich: See comments above regarding `with_` prefix and "`count` typo".

tests/geo_commands_test.py
polyn0m: Added *_binary tests

tests/geo_commands_test.py
polyn0m: Added test_geo_not_exist_members test

popravich:
Good job! Thanks

polyn0m:
You are welcome

PR #179, opened 2017-01-06
Few enhancements to Geo commands

popravich:
changes:
* [x] Rename `GeoCoord` to `GeoPoint`.
* [x] Add mixin to docs.
* [x] `georadius`/`georadiusbymember` parameters after `unit` made keyword-only.


popravich:
cc: @polyn0m few changes to API

polyn0m:
I think later need made async-for syntax support for `georadius` / `georadiusbymember`, because it's based on sorted set in Redid.

PR #191, opened 2017-01-26
Fix raising of deprecation warning

Jamim:
Pool creating using deprecated _commands_factory_ argument fails with the following error:
`AttributeError: module 'warnings' has no attribute 'DeprecationWarning'`


Jamim:
Also, it looks like there is an issue with CI.
Build for Python 3.4 failed for some reason that seems unrelated to this fix.

popravich:
Thanks

PR #192, opened 2017-01-26
Avoid deprecation warnings due to legacy __aiter__ protocol.

the-panda:
Starting with Python 3.5.2 the coroutine `__aiter__` functions will throw PendingDeprecationWarnings. Starting with Python 3.6 it will be actual DeprecationWarnings.

See https://www.python.org/dev/peps/pep-0492/#api-design-and-implementation-revisions

### DIFF
--- a/aioredis/pubsub.py
+++ b/aioredis/pubsub.py
@@ -4,7 +4,7 @@ import sys
 import types
 
 from .abc import AbcChannel
-from .util import create_future, _converters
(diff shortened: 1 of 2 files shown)


the-panda:
Sorry for the decreased coverage, but in fact all new lines are covered by the tests (but not in the same version of Python).

popravich:
Thanks

PR #198, opened 2017-02-19
Update appveyor hiredis install.

### LINKED ISSUE #197: Change the hiredis install method.
A newer version of hiredis is available on github not pip. It would have to be installed lik this instead:

```cmd
pip install --upgrade https://github.com/redis/hiredis-py/archive/master.zip
```

Hope you understand that the one on PyPi was so old that it is legit 2 years ago. So the github one is more or so along the lines of current.

AraHaan:
Update building on appveyor to install latest hiredis.

Also will look at travis and update that to all in this one PR.

Merging this will close #197.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,7 +12,7 @@ environment:
 
 install:
 - "cinst redis-64"
-- "%PYTHON%\\python.exe -m pip install hiredis"
+- "%PYTHON%\\python.exe -m pip install --upgrade https://github.com/redis/hiredis-py/archive/master.zip"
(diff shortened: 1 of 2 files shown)


popravich:
Hi, @AraHaan!
I've checked [hiredis diff](https://github.com/redis/hiredis-py/compare/v0.2.0...HEAD) and its only few
minor changes so we can consider pypi version as "fresh").
Also I don't like running tests against unkown dependency version as it is possible that
two parallel build could run against different hiredis commits and it would be hard to replay such builds.

AraHaan:
hmm

blueyed:
I agree with @popravich.

Let's close this?

PR #205, opened 2017-04-13
backward compatability with `yield from as conn`

zemlanin:
https://github.com/aio-libs/aiohttp-session/blob/fe21ec8a68e8b1030dade1a363f9c0da6495ad1c/aiohttp_session/redis_storage.py#L28


popravich:
Thanks

PR #207, opened 2017-04-22
Add asyncio trove classifier

alefteris:
The `Framework :: AsyncIO` [trove classifier](https://pypi.python.org/pypi?%3Aaction=list_classifiers) has recently been added to PyPI. It can help people discover asyncio related packages.


popravich:
Thanks

PR #210, opened 2017-04-28
Do not return a connection with pending requests to the pool

hubo1016:
A redis connection cannot return to a pool if it has pending requests, especially blocking commands like "BLPOP/BRPOP". The command may block for an indefinite time, and stops the following commands from executing if the connection is acquired from the pool. Also, because the command is still executing, the BRPOP command may pop an item and drop it which creates bugs which are very hard to debug.

### DIFF
--- a/aioredis/pool.py
+++ b/aioredis/pool.py
@@ -340,6 +340,10 @@ class ConnectionsPool(AbcPool):
                 logger.warning(
                     "Connection %r is in subscribe mode, closing it.", conn)
                 conn.close()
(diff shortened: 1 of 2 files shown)


popravich:
Thanks for the patch.
Please add some tests

hubo1016:
@popravich tests done. Some tests are failing on Travis-CI, but they are not likely related to the changes.

popravich:
Ok, thanks, I will check everything

popravich:
Thanks

PR #212, opened 2017-04-29
Python parser

popravich:
pure-python parser (PyPy3 support as a bonus)


aioredis/connection.py
```diff
@@ -107,8 +107,8 @@ def __init__(self, reader, writer, *, address, encoding=None, loop=None):
         self._address = address
         self._loop = loop
         self._waiters = deque()
-        self._parser = hiredis.Reader(protocolError=ProtocolError,
-                                      replyError=ReplyError)
+        self._parser = Reader(protocolError=ProtocolError,
```
kxepal: Why not to add a parser argument for this? And this will let you remove ugly try/except at the parser module bottom.

popravich: Yes, I will add an argument for parser class.
However hiredis will still be a requirement and default parser.
I've added PyReader to make lib run on PyPy3.

aioredis/parser.py
```diff
@@ -0,0 +1,161 @@
...
+            raise
+        else:
+            return False
+
+
+try:
+    import hiredis
+    Reader = hiredis.Reader
+except ImportError:
+    Reader = PyReader
```
kxepal: Actually, this will never works since hiredis is in install requirements and will be always available.

popravich: if only `install_requires=['hiredis; platform_implementation_name=cpython']` worked since python3.3...
there will be a hack (if statement) in setup.py to require hiredis only for `cpython` platform,
thus allowing pypy3.
I'm now trying to setup travis-ci with fresh pypy3...

aioredis/connection.py
kxepal: Good!

aioredis/parser.py
kxepal: Ah, I see. That missed bit changes whole the picture and makes sense.

PR #218, opened 2017-05-05
Add `parser` argument

popravich:
Allow to set custom protocol reader class.

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -43,7 +43,7 @@ _PUBSUB_COMMANDS = (
 
 @asyncio.coroutine
 def create_connection(address, *, db=None, password=None, ssl=None,
-                      encoding=None, loop=None):
+                      encoding=None, parser=None, loop=None):
     """Creates redis connection.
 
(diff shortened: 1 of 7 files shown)


aioredis/connection.py
```diff
@@ -99,15 +104,20 @@ def create_connection(address, *, db=None, password=None, ssl=None,
...
     """Redis connection."""
 
-    def __init__(self, reader, writer, *, address, encoding=None, loop=None):
+    def __init__(self, reader, writer, *, address, encoding=None,
+                 parser=None, loop=None):
         if loop is None:
             loop = asyncio.get_event_loop()
+        if parser is None:
+            parser = Reader
+        assert callable(parser), (
```
kxepal: May be check against some interface implementation? Or doesn't worth for now?

popravich: well, it's more like a fool-proof. I think we can add proper check later, if needed...

kxepal: Ok.

Issue #184, opened 2017-01-17
Add connect timeout


popravich:
aioredis may stuck (pool especially) if Redis server is not responsive.

(don't know why nobody notice it is missing)

popravich:
TODO:
* [ ] add connect timeout
* [ ] test with slow connection

SeavantUUz:
I met some problem, but I could not support your suggestion. What I had seen, although redis server is fine, aioredis got stuck too.

In redis-py, why you could pass an optional argument to pool, is redis-py chose block queue as its pool's implement. The timeout argument actually is used to tell pool the queue had been empty for specific time. aioredis used deque as its pool's implement, so you can't pass a timeout.

Another time out argument is `socket_timeout` or `socket_keepalive` and so on. I'm not sure whether they are good choices.

popravich:
What I mean is that aioredis can stuck for some time (probably long enough) when using pool,
basically, the call to `ConnectionsPool._create_new_connection()` need to have some timeout
otherwise it will block the whole pool.

pfreixes:
We might be having this issue, I can work on a fix ASAP

pfreixes:
This can be closed.

PR #221, opened 2017-05-10
Support for connection timeout param

pfreixes:
Related to #184, it allows aioredis to configure a limited
time that will be used trying to open a connection if it is
reached an `asyncio.TimeoutError` will be raised

By default, any timeout is configured.

Not very happy with the current test implementation, at integration level there is not an easy way to test it, if you come with something better please ping me.

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -72,9 +72,13 @@ def create_connection(address, *, db=None, password=None, ssl=None,
     """
     assert isinstance(address, (tuple, list, str)), "tuple or str expected"
 
+    if timeout is not None and timeout <= 0:
+        raise ValueError("Timeout has to be None or a number greater than 0")
+
     if isinstance(address, (list, tuple)):
         host, port = address
         logger.debug("Creating tcp connection to %r", address)
+        print(asyncio.open_connection)
         reader, writer = yield from asyncio.wait_for(asyncio.open_connection(
             host, port, ssl=ssl, loop=loop), timeout, loop=loop)
         sock = writer.transport.get_extra_info('socket')
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -78,7 +78,6 @@ def create_connection(address, *, db=None, password=None, ssl=None,
     if isinstance(address, (list, tuple)):
         host, port = address
         logger.debug("Creating tcp connection to %r", address)
-        print(asyncio.open_connection)
         reader, writer = yield from asyncio.wait_for(asyncio.open_connection(
             host, port, ssl=ssl, loop=loop), timeout, loop=loop)
         sock = writer.transport.get_extra_info('socket')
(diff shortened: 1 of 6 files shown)


aioredis/pool.py
```diff
@@ -17,7 +17,7 @@
 @asyncio.coroutine
 def create_pool(address, *, db=None, password=None, ssl=None, encoding=None,
                 minsize=1, maxsize=10, commands_factory=_NOTSET,
-                parser=None, loop=None):
+                parser=None, loop=None, timeout_create_connection=None):
```
popravich: Lets rename this argument to `create_connection_timeout`

aioredis/pool.py
```diff
@@ -392,13 +396,15 @@ def _fill_free(self, *, override_min):
                     # connection may be closed at yield point
                     self._drop_closed()
 
+    @asyncio.coroutine
```
popravich: no need for this decorator here

tests/connection_test.py
```diff
@@ -33,6 +33,13 @@ def test_connect_tcp(request, create_connection, loop, server):
 
 
 @pytest.mark.run_loop
+def test_connect_tcp_timeout(request, create_connection, loop, server):
+    with pytest.raises(asyncio.TimeoutError):
+        yield from create_connection(
+            server.tcp_address, loop=loop, timeout=0)
```
popravich: Zero timeout does not make much sense — cancelling connection on next loop iteration.
I think we should check if `timeout` is `None` or `>= 0`, if timeout is 0 replace it with `None`,
otherwise (if < 0) throw `ValueError`.

popravich: Regarding tests, I think we can patch `asyncio.open_connection` with `mock`, something like this:
```python
@mock.patch('aioredis.connection.asyncio.open_connection')
def test(open_conn_mock):
    open_conn.side_effect = lambda *a, **k: asyncio.sleep(longer_than_timeout)
    ...
```

popravich:
Thanks, for PR

pfreixes:
Changes asked applied, thanks for the reviewing.

aioredis/connection.py
```diff
@@ -68,20 +72,25 @@ def create_connection(address, *, db=None, password=None, ssl=None,
...
 
+    if timeout is not None and timeout <= 0:
+        raise ValueError("Timeout has to be None or a number greater than 0")
+
     if isinstance(address, (list, tuple)):
         host, port = address
         logger.debug("Creating tcp connection to %r", address)
-        reader, writer = yield from asyncio.open_connection(
-            host, port, ssl=ssl, loop=loop)
+        print(asyncio.open_connection)
```
popravich: Please remove `print` statement.

pfreixes:
Done, thanks.

popravich:
Thanks

PR #224, opened 2017-05-12
Narrow the coverage to the aioredis package

pfreixes:
Before this commit the coverage was missleading, gathering the code
of the test aside of the aioredis package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ test:
 	$(PYTEST)
 
 cov coverage:
-	$(PYTEST) --cov
+	$(PYTEST) --cov=aioredis
 
 dist: clean man-doc
 	$(PYTHON) setup.py sdist bdist_wheel


kxepal:
I think coverage of the test suite is very important metric as well. If you have tests, functions, or branches which were not executed, then this is indeed a bloat code or, probably, a mistake which hides some real problem in design or code. Don't you think so?

pfreixes:
No sure how important are the metrics related to the test suite, but in any case, you can mix both metrics otherwise you can get a invalid impression of the coverage, either the package or the test suite.

pfreixes:
As an example take a look at the `aiohttp` coverage [1], they only cover the package.  There are other examples [2] [3]

I've found a counterexample by `aiopg` [4], looks like @asvetlov committed the change to get coverage from the main package and the tests. Maybe he wants to give his thoughts about if either the coverage has to include the tests code or not.

[1] https://codecov.io/gh/aio-libs/aiohttp
[2] https://github.com/pallets/flask/blob/master/tox.ini#L14
[3] https://github.com/channelcat/sanic/blob/master/.coveragerc
[4] https://github.com/aio-libs/aiopg/blob/master/.coveragerc

jettify:
Agree with @kxepal , coverage number is not important metric for tests by itself, important part is to see what tests have never been executed and that may hide important problems in the code.

popravich:
First of all this changes do not affect coverage collected by codecov and shown in the badge, because included pathes are configured in `setup.cfg`.
Detailed coverage is also available at [codecov.io](https://codecov.io/gh/aio-libs/aioredis) (with split by package itself and tests).
I personally against disabling tests coverage as it helps to see if tests are run as expected,
especially when you have many conditions determining test execution flow.
It helps much more than affects coverage percentage.

pfreixes:
Nothing to say if the coverage percentage btw tests and the main project can be viewed separately.

I can understand your thoughts about having the tests inside of the coverage, but the benefits of having them are out of the scope of the goal of the code coverage that must inform to ... _measure used to describe the degree to which the source code of a program is tested by a particular test suite_ [1], and it should reflect a metric that has to be primarily useful for the user.

If there are no more comments this PR can be closed.

[1] https://en.wikipedia.org/wiki/Code_coverage

PR #239, opened 2017-06-05
close_waiter is lazily created

argaen:


aioredis/pool.py
```diff
@@ -160,6 +161,8 @@ def closed(self):
     @asyncio.coroutine
     def wait_closed(self):
         """Wait until pool gets closed."""
+        if self._close_waiter is None:
+            self._close_waiter = async_task(self._do_close(), loop=self._loop)
```
popravich: I think this method can be simplified to look like this
```python
@asyncio.coroutine
def wait_closed(self):
    yield from self._close_state.wait()
    assert self._close_waiter is not None
    yield from asyncio.shield(self._close_waiter, loop=self._loop)
```

popravich: Otherwise you should add check `if self._close_waiter is None` in `close()` method.
This also applies to sentinel pool.

argaen: I don't get your point there.

If `close` is called first, it will create the `self._close_waiter` task

If `wait_closed` is called first, it will create the self._close_waiter` and then wait until `close_state` is set

popravich: ...two tasks will try to close pool

argaen: ah True that, I'll update the code

popravich:
Thanks

Issue #231, opened 2017-05-16
When load testing the pools with timeouts, once the test is done the pool doesn't recover and keeps failing


argaen:
Hi, after an issue that was opened in https://github.com/argaen/aiocache/issues/196 I've been investigating concurrency issues with both aiomcache and aioredis.

TLDR: When load testing the pools with timeouts, once the test is done the pool doesn't recover and keeps failing.

I've been working on aiomcache implementation to fix this https://github.com/aio-libs/aiomcache/pull/46 and performs good enough. Amount of 200s is really high and server keeps working after the load test `ab -n 5000 -c 100 http://127.0.0.1:8080/`.

Here are some numbers:

**aiomcache**
```python
Concurrency Level:      100
Time taken for tests:   5.345 seconds
Complete requests:      5000
Failed requests:        0
Total transferred:      935000 bytes
HTML transferred:       180000 bytes
Requests per second:    935.44 [#/sec] (mean)
Time per request:       106.901 [ms] (mean)
Time per request:       1.069 [ms] (mean, across all concurrent requests)
Transfer rate:          170.83 [Kbytes/sec] received
```

**aioredis**
```python
Concurrency Level:      100
Time taken for tests:   10.578 seconds
Complete requests:      5000
Failed requests:        4843
   (Connect: 0, Receive: 0, Length: 4843, Exceptions: 0)
Non-2xx responses:      4843
Total transferred:      741280 bytes
HTML transferred:       5652 bytes
Requests per second:    472.68 [#/sec] (mean)
Time per request:       211.559 [ms] (mean)
Time per request:       2.116 [ms] (mean, across all concurrent requests)
Transfer rate:          68.44 [Kbytes/sec] received
```

Even worse, in my case all requests keep failing when the test has finished.

This is the script I've been using for the load testing:

```python
import uuid
import logging
import asyncio
import aiocache

from aiohttp import web

logger = logging.getLogger(__name__)


class CacheManager:
    def __init__(self):
        # self.cache = aiocache.MemcachedCache(pool_size=4)
        self.cache = aiocache.RedisCache(pool_max_size=4)

    async def get(self, key):
        return await self.cache.get(key, timeout=0.1)

    async def set(self, key, value):
        return await self.cache.set(key, value, timeout=0.1)


async def handler_get(req):
    try:
        data = await req.app['cache'].get('testkey')
        if data:
            return web.Response(text=data)
        data = str(uuid.uuid4())
[...]

PR #241, opened 2017-06-07
#231 Patch for the cpython issue with Lock bpo-27585

pfreixes:
Copied and adapted code from the PR pending to be merged yet [1]

[1] https://github.com/python/cpython/pull/1031

### DIFF
--- a/aioredis/locks.py
+++ b/aioredis/locks.py
@@ -13,7 +13,7 @@ from asyncio import futures
 
 class Lock(_Lock):
 
-    if sys.version_info < (3, 5):
+    if sys.version_info < (3, 5):  # pragma: no cover
         @coroutine
         def acquire(self):
             """Acquire a lock.
@@ -55,7 +55,7 @@ class Lock(_Lock):
                 self._locked = True
                 return True
             except futures.CancelledError:
-                if not self._locked:
+                if not self._locked:  # pragma: no cover
                     self._wake_up_first()
                 raise
             finally:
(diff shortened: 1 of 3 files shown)


pfreixes:
@popravich any interest in bake a minor release with that fix?

aioredis/locks.py
```diff
@@ -0,0 +1,68 @@
...
+                yield from fut
+                self._locked = True
+                return True
+            except futures.CancelledError:
+                if not self._locked:
+                    self._wake_up_first()
+                raise
+            finally:
+                self._waiters.remove(fut)
+    else:
```
popravich: Why do you need two versions of Lock?

popravich: *Lock.acquire

pfreixes: since Python 3.5 there were some small changes in that function, I've preferred to keep both versions separately to make it readable and of course, pass the tests with all Python 3 versions.

popravich: well, that changes were actually fixing another cancellation issue https://github.com/python/asyncio/pull/393

pfreixes: yes, a shame of me. Take a look to the last commit, unified to a unique function that should be compatible with all Python versions and free of issues.

pfreixes:
what about a minor release with that fix?

argaen:
+1, also interested on trying a version with the new fixes :)

popravich:
In the middle of a process, give me a moment)

popravich:
New versions (`v1.0.0b1` and `v0.3.2`) have been released.

pfreixes:
thanks !

PR #248, opened 2017-06-16
Increase coverage for redis commands server class

pfreixes:
For those methods that can't be tested in a deterministic way using a real connection Redis, just patched and asserting the calls made to the underlying `execute` function.

It should help to pass PR like this https://github.com/aio-libs/aioredis/pull/245

### DIFF
--- a/aioredis/commands/server.py
+++ b/aioredis/commands/server.py
@@ -16,12 +16,18 @@ class ServerCommandsMixin:
     def bgrewriteaof(self):
         """Asynchronously rewrite the append-only file."""
         fut = self.execute(b'BGREWRITEAOF')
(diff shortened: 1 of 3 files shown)


pfreixes:
an easy one @popravich

blueyed:
Can you resolve the conflicts, please?

waketzheng:
@blueyed It seems this PR should be closed now. 

After comparing aredis, asyncio-redis and aioredis, we chose this one for `Sanic` projects and they have been running well in production environment for more than one year. 
Now I wanted contribute some code for this library. Before do that, I read the openning pull requests to aviod repeat. 
I think it is better to reduce outdate issues and pull requests.

Issue #186, opened 2017-01-18
Fix "ERR max number of clients reached" on connect


popravich:
When creating connection to Redis instance with clients limit reached
`ERR max number of clients reached` message is sent.
`_read_data` task will receive it however there would be no `_waiters` to process it —
no command issues / not in pub/sub mode.
Currently there is an assertion on non-empty `_waiters` list.

TODO:
* [ ] raise MaxClientsError (or any of ReplyError, RedisError);
* [ ] keep the assertion on non-empty `_waiters` list to track other possible cases
       when messages received before command sent.

pfreixes:
Im gonna work on that if you don't mind

pfreixes:
I have a proposal, just catch the error and reraise it at first attempt to send a command. The connection is also closed. Just a POC, thoughts?

```diff
diff --git a/aioredis/connection.py b/aioredis/connection.py
index b31c886..adf251d 100644
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -23,6 +23,7 @@ from .errors import (
     ReplyError,
     WatchVariableError,
     ReadOnlyError,
+    MaxClientsError
     )
 from .pubsub import Channel
 from .abc import AbcChannel
@@ -150,6 +151,7 @@ class RedisConnection(AbcConnection):
         self._pubsub_channels = coerced_keys_dict()
         self._pubsub_patterns = coerced_keys_dict()
         self._encoding = encoding
+        self._exc_before_command = None

     def __repr__(self):
         return '<RedisConnection [db:{}]>'.format(self._db)
@@ -167,28 +169,34 @@ class RedisConnection(AbcConnection):
                 #       before response
                 logger.error("Exception on data read %r", exc, exc_info=True)
                 break
+
             if data == b'' and self._reader.at_eof():
                 logger.debug("Connection has been closed by server")
                 break
-            self._parser.feed(data)
-            while True:
-                try:
-                    obj = self._parser.gets()
-                except ProtocolError as exc:
-                    # ProtocolError is fatal
-                    # so connection must be closed
-                    if self._in_transaction is not None:
-                        self._transaction_error = exc
-                    self._closing = True
-                    self._do_close(exc)
-                    return
-                else:
-                    if obj is False:
-                        break
-                    if self._in_pubsub:
-                        self._process_pubsub(obj)
+            elif data == b'-ERR max number of clients reached\r\n':
+                self._exc_before_command = MaxClientsError()
+                break
+            else:
[...]

PR #262, opened 2017-07-04
#186 Support for MaxClientsError for connection and pool

pfreixes:
When a new connection tries to get connected to the redis server and it
has been configured with a maximum number of connections, the client
will get a `MaxClientsError` exception if this limit is reached out.

Also if a pool initialization with a `min_size` that cant meet the
requirements - beucase of that limit u others issues - the pool creation
will return also a `MaxClientsError`.

### DIFF
--- a/aioredis/pool.py
+++ b/aioredis/pool.py
@@ -394,6 +394,9 @@ class ConnectionsPool(AbcPool):
             self._acquiring += 1
             try:
                 conn = yield from self._create_new_connection(address)
+                # check the healthy of that connection, if
+                # something went wrong just trigger the Exception
+                yield from conn.execute('ping')
                 self._pool.append(conn)
             finally:
                 self._acquiring -= 1
--- a/aioredis/__init__.py
+++ b/aioredis/__init__.py
@@ -16,6 +16,7 @@ from .errors import (
     ReadOnlyError,
     RedisError,
     ReplyError,
+    MaxClientsError,
     ChannelClosedError,
     WatchVariableError,
     PoolClosedError,
@@ -43,6 +44,7 @@ __all__ = [
     # Errors
     'RedisError',
     'ReplyError',
+    'MaxClientsError',
     'ProtocolError',
     'PipelineError',
     'MultiExecError',
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -23,6 +23,7 @@ from .errors import (
     ReplyError,
     WatchVariableError,
     ReadOnlyError,
+    MaxClientsError
     )
 from .pubsub import Channel
 from .abc import AbcChannel
@@ -139,6 +140,7 @@ class RedisConnection(AbcConnection):
         self._parser = parser(protocolError=ProtocolError,
                               replyError=ReplyError)
         self._reader_task = async_task(self._read_data(), loop=self._loop)
+        self._close_msg = None
         self._db = 0
         self._closing = False
         self._closed = False
@@ -185,6 +187,11 @@ class RedisConnection(AbcConnection):
                 else:
                     if obj is False:
                         break
+                    elif isinstance(obj, MaxClientsError):
+                        self._closing = True
+                        self._do_close(obj)
+                        return
+
                     if self._in_pubsub:
                         self._process_pubsub(obj)
                     else:
@@ -257,7 +264,9 @@ class RedisConnection(AbcConnection):
           is broken.
         """
         if self._reader is None or self._reader.at_eof():
-            raise ConnectionClosedError("Connection closed or corrupted")
+            msg = self._close_msg if self._close_msg else\
+                    "Connection closed or corrupted"
+            raise ConnectionClosedError(msg)
         if command is None:
             raise TypeError("command must not be None")
         if None in set(args):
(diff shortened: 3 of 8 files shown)


aioredis/pool.py
```diff
@@ -394,6 +394,9 @@ def _fill_free(self, *, override_min):
             self._acquiring += 1
             try:
                 conn = yield from self._create_new_connection(address)
+                # check the healthy of that connection, if
+                # something went wrong just trigger the Exception
+                yield from conn.execute('ping')
```
pfreixes: No really happy with that change - to much implicit -, but it is needed a deterministic way to check that the pool can be created with that min_size requirements.  Should we keep this way to check that a connection is not really opened or try something more sophisticated (hard problem)?

Also have in mind that the code that is above when the kw `override_min` is True, should be following the same pattern than here?

pfreixes:
Looks like the test suite executed by appveryor has raised an interesting buggy behavior with asyncio. If there are no waiters waiting in a stream and the stream gets first some data and later a close connection from the other peer, with the current implementation the client wont be able to grab the buffer data that was sent first. Having always an exception notifying that the connection was closed. IMHO a bug in asyncio.

Articulate a fix in `aioredis` might be complicated, but the scenario that happens in the appveyor environment - Windows? - can happen also in other systems. It's just a matter of how the code is implemented, indeed I was able to reproduce it easily. Therefore, my first thoughts will go to implement the hack.

BTW I will raise this conversation to the Python async community to get feedback, I will work with a fix ASAP .

popravich:
Oh, that `CancelledError` is misleading, we should change that.
`_read_data` coroutine has this block with comment:
```python
        while not self._reader.at_eof():
            try:
                data = yield from self._reader.read(MAX_CHUNK_SIZE)
            except asyncio.CancelledError:
                break
            except Exception as exc:
                # XXX: for QUIT command connection error can be received
                #       before response
                logger.error("Exception on data read %r", exc, exc_info=True)
                break
            if data == b'' and self._reader.at_eof():
                logger.debug("Connection has been closed by server")
                break
```
I think we should do the same here as we do with MaxClientsError — close connection with exception.
We will close connection with other error (`ConnectionResetError` instead of `MaxClientsError`) but
at least not dummy `CancelledError`.
```diff
diff --git a/aioredis/connection.py b/aioredis/connection.py
index b31c886..65f6746 100644
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -166,9 +166,12 @@ class RedisConnection(AbcConnection):
                 # XXX: for QUIT command connection error can be received
                 #       before response
                 logger.error("Exception on data read %r", exc, exc_info=True)
-                break
+                self._closing = True
+                self._do_close(exc)
+                return
             if data == b'' and self._reader.at_eof():
                 logger.debug("Connection has been closed by server")
+                # NOTE: probably set ConnectionClosedError here as well.
                 break
             self._parser.feed(data)
             while True:
```
Your thoughts?

pfreixes:
Correct me if I'm wrong but I would say that this piece of code is something that is really coupled to the transaction command. Trying to make happy all of the paths when one of the commands belonging to a transaction closes the connection. If you change that code the following test will start having issues [1].

My 2cents here. The issue that I've commented before is still possible [2]. And the way to catch them is the piece of code that you commented. But we have two problems here:

1) The `maxclients` test might become a heisentest. Sometimes parsing the `ReplyError` and sometimes not. This is what is happening right now between `appveyor` and `travis`. Might be Im wrong, but in any case, this is a plausible scenario.
2) The code that we should touch has strong dependencies with the transaction behavior and how the pending futures must be closed once the Redis connection has been closed.


[1]  https://github.com/aio-libs/aioredis/blob/master/tests/transaction_commands_test.py#L60
[2] http://bugs.python.org/issue30861

pfreixes:
Looks like that the bug [1] is moving ahead in somehow, but obviously it won't get rid of the pain taking into account the multiple python versions that aioredis supports.

I'm a bit stuck here, not sure how to tackle this PR having in mind the many factors that are involved:

- Can't be possible with the current StreamReader implementation gather the buffer if the RST  already came into the OS.
- Dependencies with the transaction command. 

Any thoughts on how move ahead with all of this?

[1] http://bugs.python.org/issue30861

popravich:
Here are my quick thoughts:
* this PR fixes *max clients* issue;
* with another PR we can replace asyncio.StreamReader with our own.
This way we can solve this RST case and also this can make `aioredis` a bit faster,
instead of copying data into `StreamReader`'s buffer then moving it into `hiredis`' Reader
we can send it directly to hiredis (although must consider hiredis buffer limits).

pfreixes:
Then if you don't mind let's keep this PR frozen till we are able to provide our own `StreamReader`. 

So once this is implemented and merged, this PR would get the pipeline in green automatically. Even though my gut feeling says that the transaction command will give us some headaches.

I will try to work with the `StreamReader` ASAP, or do you wanna take it?

popravich:
> I will try to work with the StreamReader ASAP, or do you wanna take it?

Go ahead with this, I don't have much free time for now(

popravich:
Rebased and merged via #325

PR #264, opened 2017-07-06
Support PING in pubsub mode

thomasst:
Fixes https://github.com/aio-libs/aioredis/issues/249

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -39,7 +39,6 @@ _PUBSUB_COMMANDS = (
     'PSUBSCRIBE', b'PSUBSCRIBE',
     'UNSUBSCRIBE', b'UNSUBSCRIBE',
     'PUNSUBSCRIBE', b'PUNSUBSCRIBE',
-    'PING', b'PING',
     )
 
 
@@ -250,7 +249,7 @@ class RedisConnection(AbcConnection):
             self._pubsub_patterns[pattern].put_nowait((chan, data))
         elif kind == b'pong':
             if process_waiters and self._in_pubsub and self._waiters:
-                self._process_data(obj)
+                self._process_data(b'PONG')
         else:
             logger.warning("Unknown pubsub message received %r", obj)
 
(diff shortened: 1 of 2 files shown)


aioredis/connection.py
```diff
@@ -39,6 +39,7 @@
     'PSUBSCRIBE', b'PSUBSCRIBE',
     'UNSUBSCRIBE', b'UNSUBSCRIBE',
     'PUNSUBSCRIBE', b'PUNSUBSCRIBE',
+    'PING', b'PING',
```
popravich: Try removing `ping` from `_PUBSUB_COMMANDS`, that is the reason why pool tests are failing

thomasst: This will fail then though:

```
        is_pubsub = command in _PUBSUB_COMMANDS
        if self._in_pubsub and not is_pubsub:
>           raise RedisError("Connection in SUBSCRIBE mode")
E           aioredis.errors.RedisError: Connection in SUBSCRIBE mode

aioredis/connection.py:274: RedisError
```

popravich: Yes, it will fail.
You can extend condition of that `if` to something like:
```python
if self._in_pubsub and not (is_pubsub or is_ping):
    ...
```

thomasst:
Just pushed some fixes. I also wanted to get your opinion on the return value of `ping()`. Normally it returns the string "PONG" (which Redis returns). However, in pubsub mode is returns a (lower case) "pong" message, so for consistency reasons I chose to return the "PONG" string. We might also want to consider returning nothing or True (the regular Python Redis library returns True). What do you think?

thomasst:
And do you know why Travis tests fail with `AssertionError: (<class 'aioredis.errors.ReplyError'>, ReplyError('ERR max number of clients reached',))`?

thomasst:
@popravich could you please take another look at this PR? Thanks!

popravich:
Thanks,
Sorry for long delay

Issue #271, opened 2017-07-20
Implement custom StreamReader


popravich:
Rationale:
* Get rid of bouble buffering (`data from socket -> asyncio.StreamReader.buffer -> hiredis.Reader.buffer`);
  our StreamReader can feed data directly to parser (hiredis or PyReader);
* Step over this issue — http://bugs.python.org/issue30861;

Some specs:
* must inherit from `asyncio.StreamReader`;
* `read...`-family functions should probably raise an error (RuntimeError);
* provide `async def read_response()` method waiting and returning either valid response
or raising an error (protocol error, canceled, closed, etc) (maybe some other method name);

pfreixes:
Challenge accepted.

pfreixes:
Just one note about that task, the implementation of that new Reader class **won't get rid of the double buffer flow**. In any case, the new implementation will replace the official class provided by asyncio for a new one but the double step must remain there for many reasons.

are we in the same page?

Further, we could think in implement the Reader class - will become a bunch of code lines - in Cython. Theoretically, in an environment with `uvloop` and `hiredis` must speed up `aioredis`.

pfreixes:
Forget it, we will remove the double buffer and one coroutine step.

PR #273, opened 2017-07-25
Use a custom StreamrReader #271

pfreixes:
The connection opened with the server uses a customized StreamReader
that get rid of the double buffering between the operating system and
the Redis parser. This also allows _aioredis_ to get rid of a current bug
with abruptly terminated connections.

Current implementation does not allow to keep giving support with
pause/resume the transport if some watermark has been overflowed,
the current implementation of hiredis does not publish the length of the
buffer.

Worth mentioning that doing a micro benchmark [1] the performance gains
observed is a 5%. This has been tested only with the default Asyncio loop. IMHO will worth it a discussion if a _cython_ implementation will help for environments with _uvloop_ ( I think so ).

[1] https://gist.github.com/pfreixes/e320b6f398aa7644c19188ccf2f443b1

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -16,6 +16,7 @@ from .util import (
     create_future,
     )
 from .parser import Reader
+from .stream import StreamReader
 from .errors import (
     ConnectionClosedError,
     RedisError,
@@ -85,11 +86,17 @@ def create_connection(address, *, db=None, password=None, ssl=None,
     else:
         cls = RedisConnection
 
+    if loop is None:
+        loop = asyncio.get_event_loop()
+
     if isinstance(address, (list, tuple)):
         host, port = address
         logger.debug("Creating tcp connection to %r", address)
-        reader, writer = yield from asyncio.wait_for(asyncio.open_connection(
-            host, port, ssl=ssl, loop=loop), timeout, loop=loop)
+        reader = StreamReader(limit=MAX_CHUNK_SIZE, loop=loop)
+        protocol = asyncio.StreamReaderProtocol(reader, loop=loop)
+        transport, _ = yield from asyncio.wait_for(loop.create_connection(
+            lambda: protocol, host, port, ssl=ssl), timeout, loop=loop)
+        writer = asyncio.StreamWriter(transport, protocol, reader, loop)
         sock = writer.transport.get_extra_info('socket')
         if sock is not None:
             sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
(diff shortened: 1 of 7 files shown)


aioredis/connection.py
```diff
@@ -136,8 +146,9 @@ def __init__(self, reader, writer, *, address, encoding=None,
         self._address = address
         self._loop = loop
         self._waiters = deque()
-        self._parser = parser(protocolError=ProtocolError,
-                              replyError=ReplyError)
+        self._reader.set_parser(
+            parser(protocolError=ProtocolError, replyError=ReplyError)
+        )
```
pfreixes: For the sake of compatibility and consistency with the current connection signature, the parser is not set in the `StreamReader` constructor. But if you believe that we can break that let's get rid the `parser` parameter and let's pass as a new argument of the `StreamReader` constructor

aioredis/stream.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if not data:
+            return
+
+        self._parser.feed(data)
+        self._wakeup_waiter()
+
+        # TODO: implement pause the read. Its needed
+        #       expose the len of the buffer from hiredis
+        #       to make it possible.
```
pfreixes: We don't have visibility, at least for the `hiredis` version, of the `len` of the buffer. My plans will pass doing the proper MR to expose this property by the `hiredis-py` project, doesn't look a rocket science.  With that attribute, we can implement the proper pause/resume option.

tests/transaction_commands_test.py
```diff
@@ -61,7 +61,7 @@ def test_connection_closed(redis):
     tr = redis.multi_exec()
     fut1 = tr.quit()
     fut2 = tr.incrby('foo', 1.0)
-    fut3 = tr.connection.execute('INCRBY', 'foo', '1.0')
+    fut3 = tr.incrby('foo', 1)
```
pfreixes: My understanding here is that this code had a bug? Using the old code I had a _heisentest_.

pfreixes:
weird error with CI              
> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\aioredis.1712.stdout'

Any clues @popravich?

popravich:
Hmm...
It seems that redis server process did not shut down clearly.
I guess we can change this [line](https://github.com/aio-libs/aioredis/blob/master/tests/conftest.py#L492) to catch `OSError` instead of `FileNotFoundError`. Hopefully we won't consume whole worker's tmp space.

pfreixes:
Now weird problems in the other CI ... 

> The command "sudo tar xjf python-3.3.tar.bz2 --directory /" failed and exited with 2 during .

Can you restart the jobs?

pfreixes:
green at last ... @popravich let me know your thoughts when you have the chance.

popravich:
Yeah) give me few days

pfreixes:
Any thought @popravich ? Im trying to move also the part of hiredis [1] but the project has not a lot activity recently.

In any case, FYI I'll be out for two weeks and fewer chances to have an internet connection.


[1] https://github.com/redis/hiredis-py/pull/61

popravich:
Thanks

PR #284, opened 2017-08-13
Correct error message on sentinel discovery of master/slave with password.

azinoviev:
This will help to get ReplyError "Master X error: ERR invalid password"/"Master X error: NOAUTH Authentication required." instead of MasterNotFoundError "No master found for X" when discovering password protected master/slave via Sentinel without password (or with wrong password) provided.

Pyenv's .python-version config added to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ docs/_build/
 # Vim
 .vim
 tests/ssl
+
+# pyenv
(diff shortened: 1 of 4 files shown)


popravich:
Closing this in favor of #327 (rebased and updated)

PR #289, opened 2017-09-04
Add encoding option to sorted set

jameshilliard:
Looks like this option is missing for sorted sets.


popravich:
Please, add tests for the case when encoding is specified.

jameshilliard:
I've added tests, AppVeyor seems to not be picking up my changes in this PR although tests pass [here](https://ci.appveyor.com/project/jameshilliard/aioredis/build/1.0.0b16).

PR #299, opened 2017-10-07
Adding stream commands

adamcharnock:
@antirez is [in the process](http://antirez.com/news/114) of adding streams to Redis. This PR adds initial support for the ``XADD``/``XRANGE``/``XREAD`` commands in aioredis.

## Tests

I've written tests and added the *streams* Redis branch to the travis config. In theory, the tests will only be run when the streams branch is available. ~~The logic is a bit hacky, but I figured it will only be until streams makes it into 4.0~~ (now improved as the hacky version didn't actually work).

## Style

I'm very open to feedback on my implementation style for the new commands. I've tried to stick to the aioredis style, but I think I will have probably wandered here or there. Happy to make any corrections or alterations.

## Reason

This is my first time using aioredis. I'm working on [another project](https://github.com/adamcharnock/lightbus) in which I would like to experiment with Redis streams, plus I also need asyncio support. I therefore thought I'd contribute this here.

As I said, happy to make any changes etc.

## Remaining work

- [ ] Add support for the ``MAXLEN`` parameter in ``xadd()``
- [ ] There is a question regarding style in the source for ``xread()``


adamcharnock:
It appears to me that the ``test_pool_idle_close`` test is being a little flakey on Windows. AFAIK this is not my fault... I hope.

aioredis/commands/streams.py
```diff
@@ -0,0 +1,120 @@
...
+    not currently released.
+    """
+
+    def xadd(self, stream, fields=None):
+        """ Add a message to the specified stream
+        """
+        # TODO: Add the MAXLEN parameter
+        flattened = []
+        for k, v in fields.items():
+            flattened += [k, v]
```
barrachri: Extend?

adamcharnock: Ah yes. Changed.

barrachri:
Great @adamcharnock! Do you know where I can find the implementation details related to Redis?

Did you use this: https://github.com/redis/redis-rcp/blob/master/RCP11.md?

adamcharnock:
Hi @barrachri,

I think that is a different kind of stream, one more to do with how the redis protocol encodes data on the wire.

I used the following:

http://antirez.com/news/114

The branch can be found here:

https://github.com/antirez/redis/tree/streams

barrachri:
Thanks.

Is using `OrderedDict` really important?

I was trying to think about a use case where the order of the key=value might be important after you used `fields_to_dict`.

adamcharnock:
I used an OrderedDict simply because the ordering of the key/value pairs is explicitly respected and maintained by Redis Streams. I therefore wanted to respect this in the implementation. I expect someone will find a use for it, even if I'm not sure what it would be.

Perhaps one reads a stream record then writes it back to another stream. The new record has a different field order. As Redis guarantees order, other clients could legitimately decide that this copied record was different to the original, despite there being no intention to modify it during the copying process.

As a secondary reason, I've occasionally seen hard-to-diagnose bugs crop you due to the non-deterministic ordering of dictionaries. This normally occurs when the developer assumes a regular dict has a particular ordering, which then sporadically changes under different environments and python versions.  Given that the key/values do in fact have ordering in this case, it seemed like an easy way to avoid this issue.

Also, I suspect ignoring the ordering provided & documented by Redis would probably cause developers to walk right into this kind of bug.

barrachri:
Clarified, thanks!

barrachri:
`Streams` has been merged inside https://github.com/antirez/redis/tree/unstable.

asvetlov:
Should we wait for redis stable with streaming feature?

adamcharnock:
@barrachri That is great to hear.

@asvetlov Not sure if that was directed at me – but I have no view on the matter.

I'll see about getting ``MAXLEN`` implemented soon.

asvetlov:
My question is addressed mostly to @popravich

adamcharnock:
I've just added support for ``MAXLEN`` as well as [the](https://github.com/antirez/redis/commit/9bb18e54380250f3fb931028952379c3ab2dec29) [new](https://github.com/antirez/redis/commit/45fe1f5e0019c6745e531da3ea51eee609f15ec3) ``XREVRANGE`` command.

I based my implementation of that of ``ZRANGE`` and ``ZREVRANGE``, in which case it seemed the implementation was just copied. I therefore did the same and likewise duplicated the relevant test.

adamcharnock:
Ok, travis is happy again. I don't think the appveyor tests are to do with my changes (they failing tests seem unrelated).

aioredis/commands/streams.py
```diff
@@ -0,0 +1,136 @@
...
+        """ Add a message to the specified stream
+        """
+        args = []
+        if max_len is not None:
+            if exact_len:
+                args.extend((b'MAXLEN', max_len))
+            else:
+                args.extend((b'MAXLEN', b'~', max_len))
+
+        args.append(message_id or b'*')
```
hoh: Why not specifying `message_id=b'*'` in the function arguments ?

adamcharnock: Good point, done

adamcharnock:
Quick update - it sounds like Redis Streams will be released in Redis 5.0 within a couple of months ([ref](https://twitter.com/antirez/status/964562493665431552)).

Some additional commands have been added. I expect to add these to this pull request once the available commands have been finalised.

adamcharnock:
I've now added support for the consumer group commands + tests. 

However, the following are not yet implemented in Redis. I've therefore set the tests to skip for now:

* `XGROUP SETID`
* `XGROUP DELGROUP`
* `XGROUP DELCONSUMER`

Also, some of the commands are not implemented entirely as documented. This'll probably require a little tweaking once streams is stable.

adamcharnock:
Update: Redis 5 RC3 is now out. What is the view on merging at this time?

adamcharnock:
Ok, travis is happy now. The code coverage diff seems to show lack of coverage on sentinel as the sticking point, but as I haven't touched that I'm going to assuming is not my fault.

aioredis/commands/streams.py
```diff
@@ -0,0 +1,222 @@
+import sys
+from collections import OrderedDict
+
+from aioredis.util import wait_convert, wait_make_dict, wait_ok
+
+PY_VER = sys.version_info
+
+if PY_VER < (3, 0):
+    from itertools import izip as zip
```
popravich: These lines don't make sense. We don't support Python 2.x.

aioredis/commands/streams.py
```diff
@@ -0,0 +1,222 @@
...
+    parsed = []
+    for stream, messages in messages_by_stream:
+        for message_id, fields in parse_messages(messages):
+            parsed.append((stream, message_id, fields))
+    return parsed
+
+
+def parse_lists_to_dicts(lists):
+    """ Convert [[a, 1, b, 2], ...] into [{a:1, b: 2}, ...]"""
+    def _list_to_dict(list_):
```
popravich: This function is almost the same as `fields_to_dict` except return type,
It is better to make a single function and pass `OrderedDict`/dict` types as argument.

aioredis/commands/streams.py
```diff
@@ -0,0 +1,222 @@
...
+        fut = self.execute(b'XREVRANGE', stream, start, stop, *extra)
+        return wait_convert(fut, parse_messages)
+
+    def xread(self, streams, timeout=0, count=None, latest_ids=None):
+        """Perform a blocking read on the given stream"""
+        args = self._xread(streams, timeout, count, latest_ids)
+        fut = self.execute(b'XREAD', *args)
+        return wait_convert(fut, parse_messages_by_stream)
+
+    def xread_group(self, group_name, consumer_name, streams, timeout=0,
```
popravich: Please add a docstring

popravich: ...and to 12 more methods below

aioredis/commands/streams.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def xgroup_delconsumer(self, stream, consumer_name):
+        fut = self.execute(b'XGROUP', b'DELCONSUMER', stream, consumer_name)
+        return wait_ok(fut)
+
+    def xpending(self, stream, group_name, start=None, stop=None, count=None,
+                 consumer=None):
+        # Returns: total pel messages, min id, max id, count
+        ssc = [start, stop, count]
+        if any(ssc) and not all(ssc):
```
popravich: This check may trigger false-positive exceptions in case when `start=0`.
Please add test for this case.

tests/stream_commands_test.py
```diff
@@ -0,0 +1,574 @@
...
+        (b'field2', b'value2')]
+    )
+
+
+@pytest.mark.run_loop
+@pytest.redis_version(999, 999, 999, reason="Streams only available on redis "
+                                            "unstable branch")
+async def test_xadd_maxlen_exact(redis, server_bin):
+    message_id1 = await redis.xadd('test_stream', {'f1': 'v1'})  # noqa
+    sleep(0.001)  # Ensure the millisecond-based message ID increments
```
popravich: Do you really need a thread-blocking sleep?

adamcharnock:
Thank you for the comments @popravich, I'll take a look at making the changes now/shortly.

aioredis/commands/streams.py
adamcharnock: Yep, not sure what I was thinking here

aioredis/commands/streams.py
adamcharnock: Ok, will do

aioredis/commands/streams.py
adamcharnock: Good catch! Will add test and fix

tests/stream_commands_test.py
adamcharnock: Ha, I certainly don't :-)

adamcharnock:
I've just pushed changes for everything bar the doc strings, which I'll take a look at now.

aioredis/commands/streams.py
```diff
@@ -0,0 +1,213 @@
+from collections import OrderedDict
+
+from aioredis.util import wait_convert, wait_make_dict, wait_ok
+
+
+def fields_to_dict(fields, type_=OrderedDict):
+    """Convert a flat list of key/values into an OrderedDict"""
+    fields_iterator = iter(fields)
+    return OrderedDict(zip(fields_iterator, fields_iterator))
```
popravich: `type_` argument is not used

adamcharnock: Urgh, thanks. Fixed.

aioredis/commands/streams.py
adamcharnock: Done now

adamcharnock:
AFAIK I've now responded to all your change requests. Do let me know if there is anything else though 👍

popravich:
Lets wait for CI tasks to complete.

popravich:
Merged! Thanks

adamcharnock:
That's great to hear, thank you very much for taking the time to review this. I'll be happy to revisit this if this API changes before Redis 5.0 stable.

Once this sees a release I'll update [lightbus](https://lightbus.org/) to point to the particular aioredis version.

Issue #306, opened 2017-10-21
undistinguishable command future cancellation


sebbov:
The following two cases are indistinguishable from a caller's perspective, when using a reconnecting client:

1. The command future was externally (by the application, possibly a higher layer one) canceled:
https://github.com/honzasp/aioredis/blob/master/aioredis/connection.py#L169

2. The read had an otherwise unhandled exception (incl in our case an unexpected, externally (network/server) initiated connection reset, `ConnectionResetError` exception):
https://github.com/honzasp/aioredis/blob/master/aioredis/connection.py#L179
(The XXX may indicate remaining work in this area that concerns the issue at hand.)
In this case, the connection ends up being closed and the waiters / command futures cancelled (whereas the expectation here would be to have a non-`asyncio.CancelledError` set on the future).

In both cases, the future returned by `aioredis` will be observed to have been canceled.

In the case of a non-reconnecting (normal) client, a canceled future due to the above can differentiate between both 1. and 2. by looking at the `client.closed` attribute.  No such luck for a reconnecting client, which always announces `client.closed = False`. If the suggested fix above isn't doable, exposing the state of the wrapped connection (e.g. a `was_closed`) in the auto-reconnecting client would still be allow to make this differentiation, and thus provide a way out.

thanks!
-seb.

popravich:
I'm not going to fix reconnecting client, as it has been deprecated and dropped in v1.0.
However I've fixed futures cancelling on disconnects, so its now possible to distinguish between
task cancellation and connection error.
The fix is in master branch but I will backport it to v0.3 branch soon.

popravich:
Released with v0.3.5

Issue #190, opened 2017-01-24
Can't distinguish between co-routine cancellation and Redis connection broken


skbergam:
I'm trying to make my application more resilient to Redis server being unavailable.  I do BRPOP and SUBSCRIBE operations, and I would like them to handle this exceptional case by sleeping for a few seconds and simply trying the operation again.  The problem I'm running into is that I just get back a `CancelledError` whenever this happens, and I can't differentiate between that and a "legitimate" `CancelledError` that happens when I decide to cancel execution of the task intentionally from within my application.

I notice from the exception output that this line is referenced:
```
File "/xyz/lib/python3.5/site-packages/aioredis/commands/__init__.py", line 42, in execute                                                                                           │2017-01-23 19:56:35,975 E Unhandled exception in asyncio task with unknown TaskWithLogger logger (<class 'NoneType'>)
    return (yield from conn.execute(*args, **kwargs)) 
```

But that doesn't appear to match the latest code I see in `master`, but I confirmed I have the latest version from pip.

2 questions:

1. Is there any way to differentiate between these two exception cases? 
2. Does the current code in `master` behave any differently?

popravich:
Hi, @skbergam 

> 1. Is there any way to differentiate between these two exception cases?

Currently there is no way to distinguish between these two.
I am thinking on this and probably I will add distinct exception for cancelling pending waiters on connection close.
However for now the only way to differentiate these two is to catch CancelledError and to check
if connection is marked as closed:
```python
for _ in range(retry_count):
    try:
        await redis.brpop(...)
    except asyncio.CancelledError:
        if not redis.closed:
            continue
        raise
    else:
        break
```
(Ofcourse both task cancellation and connection close can happen same time....)

> 2. Does the current code in master behave any differently?

In terms of cancellation — no, it behaves differently in terms of API, see #40 and [example](http://aioredis.readthedocs.io/en/latest/examples.html#commands-example)

skbergam:
@popravich: Thank you for the help.  Unfortunately checking `.closed` in this case doesn't seem to work, as the value is always `False`  in both cases: (1. stopping the redis server to simulate a connection issue, and 2. cancelling the task myself).

Here is my code:
```
import asyncio
import aioredis
import sys
import traceback as tb

async def main_async(readline_async):
    task = asyncio.ensure_future(run_in_a_loop())
    print("waiting for BRPOP in a loop. press [ENTER] to cancel.")
    await readline_async()
    task.cancel()
    await task

[...]

PR #316, opened 2017-11-07
Distinct error for futures cancellation

popravich:
Cancel pending futures with one of distinct errors (`ConnectionClosedError`, 
`ConnectionForcedCloseError` or builtin exception like `ConnectionError`)
instead of calling `cancel()` method.


popravich:
This would resolve #306 and #190

PR #317, opened 2017-11-07
Do not install hiredis by default

fspot:
`hiredis` is already optional from the source code's point of view, but is still installled by default on cpython platform : this PR make this installation optional (`pip install aioredis[hiredis]` if you want hiredis).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,10 @@
 import re
 import os.path
 import sys
-import platform
 from setuptools import setup, find_packages
 
 
 install_requires = ['async-timeout']
-if platform.python_implementation() == 'CPython':
-    install_requires.append('hiredis')
 
 PY_VER = sys.version_info
 
@@ -67,4 +64,5 @@ setup(name='aioredis',
       packages=find_packages(exclude=["tests"]),
       install_requires=install_requires,
       include_package_data=True,
+      extras_require={'hiredis': ["hiredis"]}
       )


asvetlov:
Your intention is great but tests should be fixed before merging.

popravich:
Main idea of having hiredis as optional dependency is to be able to use aioredis on PyPy3
(and probably other non-CPython platforms). It is more like trade-off to have "+1 platform" at the cost
of parser performance.
On CPython platform most users will prefer hiredis (for its performance),
so what is your reason for making it optional?

fspot:
@popravich actually I needed this because the introduction of hiredis in my project caused an incompatibility with celery (https://github.com/celery/celery/issues/3898), so in the "+1 platforms" you can potentially include celery.

I do however understand that having hiredis by default is probably what most people want, it's sad that you cant "opt-out" of an optional dependency, like `aioredis[-hiredis]`.

popravich:
> hiredis in my project caused an incompatibility with celery

Ouch, hilarious.
>  it's sad that you cant "opt-out" of an optional dependency

You can easily install `aioredis` without `hiredis` package by just using `--no-deps` flag in `pip install`:
```bash
$ pip install --no-deps aioredis
```

fspot:
> Ouch, hilarious.

What do you mean ? It is an annoying bug which, in my case, can be solved by having the hiredis dependency optional in aioredis. I proposed this PR only because I saw there was a recent effort in the codebase made to make hiredis optional.

I did not know about the `--no-deps` flag, but in my case aioredis is one of the many dependencies listed in my `requirements.txt`, and this flag would apply to all of them :confused: . Using something like pip-tools to get an exhaustive list of dependencies + remove the hiredis line + use pip -r with `--no-deps` flag is another way to solve the problem, but making hiredis optional seemed smarter.

asvetlov:
If celery is broken -- maybe fixing it is even smarter than dropping hiredis?

fspot:
Sure, I will use my fork in the meantime. Thank you for your feedback, I close this PR.

PR #321, opened 2017-11-09
Drop support for Python 3.3 and 3.4

### LINKED ISSUE #304: Let's drop Python 3.3 & 3.4 support
Python 3.3 is too old and not maintained anymore.
I doubt if anybody runs aioredis on this version.

Honestly it's time to drop 3.4 too -- the next aiohttp release most likely will support python 3.5+ only.

hugovk:
Fixes #304.

Also drop < Python 3.5.3.


hugovk:
I've not used Python's await/async before, perhaps someone more familiar it could tackle those in another issue/PR?

popravich:
Ok, fine

PR #325, opened 2017-11-14
Support maxclients error #186

### LINKED ISSUE #186: Fix "ERR max number of clients reached" on connect

popravich:
Rebased #262;
Resolves #186 and #262 ;

### DIFF
--- a/aioredis/__init__.py
+++ b/aioredis/__init__.py
@@ -16,6 +16,7 @@ from .errors import (
     ReadOnlyError,
     RedisError,
     ReplyError,
+    MaxClientsError,
     ChannelClosedError,
     WatchVariableError,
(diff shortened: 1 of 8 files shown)


pfreixes:
Thanks for taking care of that.

PR #327, opened 2017-11-14
#284 Correct error message on sentinel discovery of master/slave with password

popravich:
Rebased from master;

Use more specific error kind.


popravich:
@azinoviev, your thoughts?

Issue #328, opened 2017-11-14
Document how to run tests locally


adamrothman:
This will make things much easier for contributors who are not already familiar with how to do so.

popravich:
http://aioredis.readthedocs.io/en/latest/devel.html

adamrothman:
Oops. 😄

PR #329, opened 2017-11-14
Support passing bytearray to Redis commands (closes #324)

### LINKED ISSUE #324: Can't pass bytearray as value to SET
Given that `bytes` can be used as arguments to Redis commands, it seems reasonable to also support `bytearray`.

This is currently impossible due to a check for `None` in the args to `RedisConnection::execute` which involves the creation of a set from said args. While I understand this optimization, it precludes the use of `bytearray` as it is not hashable (see below).

No Redis command accepts more than 10 or so args; just scanning the list without creating a set feels like an acceptable tradeoff.

## Example

Here's a snippet from an IPython session demonstrating the issue:

```
In [6]: data
Out[6]: bytearray(b'hello world')

In [7]: result = loop.run_until_complete(redis.set('some-sweet-key', data, expire=86400))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-7f1e0c5ed0b8> in <module>()
----> 1 result = loop.run_until_complete(redis.set('some-sweet-key', data, expire=86400))

/usr/local/lib/python3.6/site-packages/aioredis/commands/string.py in set(self, key, value, expire, pexpire, exist)
    184         elif exist is self.SET_IF_NOT_EXIST:
    185             args.append(b'NX')
--> 186         fut = self.execute(b'SET', key, value, *args)
    187         return wait_ok(fut)
    188 

/usr/local/lib/python3.6/site-packages/aioredis/commands/__init__.py in execute(self, command, *args, **kwargs)
     50 
     51     def execute(self, command, *args, **kwargs):
---> 52         return self._pool_or_conn.execute(command, *args, **kwargs)
     53 
     54     def close(self):

/usr/local/lib/python3.6/site-packages/aioredis/pool.py in execute(self, command, *args, **kw)
    200         conn, address = self.get_connection(command, args)
    201         if conn is not None:
--> 202             fut = conn.execute(command, *args, **kw)
    203             return self._check_result(fut, command, args, kw)
    204         else:

/usr/local/lib/python3.6/site-packages/aioredis/connection.py in execute(self, command, encoding, *args)
    261         if command is None:
    262             raise TypeError("command must not be None")
--> 263         if None in set(args):
    264             raise TypeError("args must not contain None")
    265         command = command.upper().strip()

TypeError: unhashable type: 'bytearray'
```

## Environment

* Python 3.6
* aioredis 1.0.0b2

adamrothman:
I have not actually run the tests myself as there is no documentation on how to do so (#328). But I believe these changes are sufficient to fix the issue.

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -287,7 +287,7 @@ class RedisConnection(AbcConnection):
             raise ConnectionClosedError(msg)
         if command is None:
             raise TypeError("command must not be None")
-        if None in set(args):
+        if None in args:
             raise TypeError("args must not contain None")
         command = command.upper().strip()
         is_pubsub = command in _PUBSUB_COMMANDS
--- a/aioredis/util.py
+++ b/aioredis/util.py
@@ -8,11 +8,11 @@ _NOTSET = object()
 # NOTE: never put here anything else;
 #       just this basic types
 _converters = {
-    bytearray: lambda val: val,
     bytes: lambda val: val,
-    float: lambda val: str(val).encode('utf-8'),
-    int: lambda val: str(val).encode('utf-8'),
+    bytearray: lambda val: val,
     str: lambda val: val.encode('utf-8'),
+    int: lambda val: str(val).encode('utf-8'),
+    float: lambda val: str(val).encode('utf-8'),
 }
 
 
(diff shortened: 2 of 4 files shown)


popravich:
Thanks

PR #351, opened 2017-12-07
Fixed ConnectionsPool.[pubsub_channels/pubsub_patterns]

frostoov:
Fixed bug in `commands.pubsub.PubSubCommandsMixin.subscribe` (and psubscribe). `ConnectionsPool.pubsub_channles` code was called before `ConnectionsPool._pubsub_conn` creation, and therefore `pubsub_channles`


popravich:
Thanks for the report.
I've pushed a bit simpler patch.

PR #370, opened 2018-01-25
Added async support for flushall() and flushdb()

cpwr:
Also, docstring has been updated

### DIFF
--- a/aioredis/commands/server.py
+++ b/aioredis/commands/server.py
@@ -127,7 +127,10 @@ class ServerCommandsMixin:
         return self.execute(b'DEBUG', 'SEGFAULT')  # pragma: no cover
 
     def flushall(self, async=False):
-        """Remove all keys from all databases."""
+        """
+        Remove all keys from all databases.
+        :param async: lets the entire dataset to be freed asynchronously. Defaults to False
+        """
         if async:
             fut = self.execute(b'FLUSHALL', b'ASYNC')
         else:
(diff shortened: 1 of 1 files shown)


cpwr:
@popravich Алексей, сориентируйте, пожалуйста, как исправить сие недоразумение

aioredis/commands/server.py
```diff
@@ -126,14 +126,26 @@ def debug_segfault(self, key):
         # won't test, this probably works
         return self.execute(b'DEBUG', 'SEGFAULT')  # pragma: no cover
 
-    def flushall(self):
-        """Remove all keys from all databases."""
-        fut = self.execute(b'FLUSHALL')
+    def flushall(self, async=False):
```
popravich: `async` is a reserved keyword, argument must be renamed (maybe `async_flush` or something better)
https://docs.python.org/3.7/reference/lexical_analysis.html#keywords

aioredis/commands/server.py
```diff
@@ -126,14 +126,26 @@ def debug_segfault(self, key):
...
+        """
+        Remove all keys from all databases.
+        :param async: lets the entire dataset to be freed asynchronously. Defaults to False
```
popravich: Please keep line length less than 80 characters (`make flake` will show this error).
Also add a line break after first sentence to make a separate paragraph for parameter description.

cpwr:
- Maybe I should rename parameter to `sync` (since it's not a keyword) and set it's default value to `True`?
- Should I write some new tests, or should I change existing, like `tests/generic_commands_test.py:500` to use new features?

popravich:
That's ok, thanks.

PR #389, opened 2018-03-01
fixes #388 - Added password redis_uri example to api_reference

### LINKED ISSUE #388: Document user/password usage
DBUri usage in redis seems to not be so-well documented, it would be helpful to have it documented here.
It is a standard URI with auth in it, but I've seen some co-workers having trouble understanding it and might (even if it's a kinda-corner case, as most people don't use authentication) be helpful to have an example available.

In this case, wich is a bit  confusing is the form:

redis://:password@host:port/db

(the :password bit, it makes sense as it is an empty username, but as said, might be helpful to have in the docs and has no cost :-) )

XayOn:
This would close #388


popravich:
Thanks

PR #426, opened 2018-07-19
Python 3.7

### LINKED ISSUE #425: calling close triggers deprecation warning in python 3.7
`DeprecationWarning: 'with await lock' is deprecated use 'async with lock' instead` in https://github.com/aio-libs/aioredis/blob/master/aioredis/pool.py#L146

Not sure if you officially support python 3.7 yet though.

kkaiser:
fixes #425

### DIFF
--- a/tests/pool_test.py
+++ b/tests/pool_test.py
@@ -538,7 +538,7 @@ async def test_await(create_pool, server, loop):
         server.tcp_address,
         minsize=10, loop=loop)
 
-    with await pool as conn:
+    with (await pool) as conn:
         msg = await conn.execute('echo', 'hello')
         assert msg == b'hello'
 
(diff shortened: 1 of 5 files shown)


kkaiser:
seems like sudo needs to be enabled in travis to support python 3.7: https://github.com/travis-ci/travis-ci/issues/9815#issuecomment-401756442

webknjaz:
Yes, sudo+xenial just for that job

tests/pool_test.py
```diff
@@ -538,7 +538,7 @@ class MyPool(ConnectionsPool):
         server.tcp_address,
         minsize=10, loop=loop)
 
-    with await pool as conn:
+    with (await pool) as conn:
```
webknjaz: Why not `async with`?

kkaiser: It could be rewritten to `async with pool.get() as conn`. The deprecation warning is not triggered on `ConnectionsPool` objects. Rewriting as `async with pool as conn` generates an `AttributeError` because `__aexit__` is not implemented like in `_AsyncConnectionContextManager`.

webknjaz: Well, then it's a part of the fix.

popravich:
Thanks

PR #453, opened 2018-09-07
Update supported python versions

michael-k:
- 3.7 was not mentioned everywhere
- minimal requirement is Python 3.5.3, not Python 3.5 (according to readme)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,7 +25,7 @@ Pub/Sub support                     Yes
 Sentinel support                    Yes [1]_
 Redis Cluster support               WIP
 Trollius (python 2.7)               No
-Tested CPython versions             `3.5, 3.6 <travis_>`_ [2]_
+Tested CPython versions             `3.5.3+, 3.6, 3.7 <travis_>`_ [2]_
 Tested PyPy3 versions               `5.9.0 <travis_>`_
 Tested for Redis server             `2.6, 2.8, 3.0, 3.2, 4.0 <travis_>`_
 Support for dev Redis server        through low-level API
--- a/setup.py
+++ b/setup.py
@@ -11,8 +11,8 @@ if platform.python_implementation() == 'CPython':
 
 PY_VER = sys.version_info
 
-if PY_VER < (3, 5):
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -11,8 +11,8 @@
 
 PY_VER = sys.version_info
 
-if PY_VER < (3, 5):
-    raise RuntimeError("aioredis doesn't support Python version prior 3.5")
+if PY_VER < (3, 5, 3):
+    raise RuntimeError("aioredis doesn't support Python version prior 3.5.3")
```
popravich: Why 3.5.3?
There is not strict requirement on `.3` version.
I've just run the tests on Python 3.5.2
Can you please clarify why exactly `3.5.3`?

michael-k: I've just picked it up from the readme: https://github.com/aio-libs/aioredis/blob/master/README.rst#requirements

docs/index.rst
```diff
@@ -25,7 +25,7 @@ Pub/Sub support                     Yes
 Sentinel support                    Yes [1]_
 Redis Cluster support               WIP
 Trollius (python 2.7)               No
-Tested CPython versions             `3.5, 3.6 <travis_>`_ [2]_
```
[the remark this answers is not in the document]
michael-k: I'm not familiar with SphinX and I'd also argue that it's a bit out of scope of this PR ;)
But yeah, it would be a great way to ensure both resources are in sync.

webknjaz:
Tests need to be fixed

michael-k:
> Tests need to be fixed

Unrelated to this PR; I've opened another PR to address the failing tests: aio-libs/aioredis#455

setup.py
popravich: Oh, I really don't remember what was the reason for exactly 3.5.3.
We, probably, can stick with `3.5.3`

bmerry: IIRC 3.5.3 changes `get_event_loop` so that when called from inside a task it always returns the event loop in which the task is running, even if it isn't the default event loop for the thread. I don't know if that is relevant to aioredis, but I believe it's the reason aiohttp now requires 3.5.3.

PR #456, opened 2018-09-10
Fix assertion error: AssertionError aioredis.pool in acquire

joaomoa:
the _drop_closed method should be called after calling _free_free and not in the end of the method itself.

This is because it is not guaranteed that the next line of code after calling `fill_free` will be executed immediately after `fill_free` (since it's an async method)

image

We have been running this branch on Smarkets. The server which talks to redis lives in https://live.smarkets.com/sockjs/info and everything is behaving as expected.


seandstewart:
This change has been superseded by #891.

PR #459, opened 2018-09-11
fix for sensitive logging

marcstreeter:
no longer logs password as requested in issue https://github.com/aio-libs/aioredis/issues/399


popravich:
Can you please add add a test capturing and checking this log message.
Logs can be captured like this https://github.com/aio-libs/aioredis/blob/master/tests/pool_test.py#L469-L479
and password setup can look like this https://github.com/aio-libs/aioredis/blob/master/tests/connection_test.py#L266-L291

marcstreeter:
Oops - right, I should provide the tests.  I’ll come back to it.  I hit a snag trying to figure out how I’d test if the password was included in the log.  The examples you showed me (as well as the other tests I searched through) only seem to pass the password via the kwargs and not the address string (only host and port are being passed there it seems).

popravich:
Are you talking about this — `redis://:passwd@host:1234/`?

blueyed:
Please use pytest's `caplog` for the tests.
(`pytest.log` is added by aioredis)

PR #472, opened 2018-10-08
[WIP] Improve support for streams

barrachri:
**WIP**

This is a first attempt to improve Redis Streams support

I added a `experiment.py` file inside the PR to play with it locally.


asvetlov:
@popravich have you a chance to take a look?

popravich:
@asvetlov, I guess it's not ready for review yet.
@barrachri, some description on what is improved (the flow "before" and "after") probably with some
example would be appreciated.

asvetlov:
@popravich thanks for feedback

barrachri:
Sure @popravich!

so right now if you want to use Redis streams you would have to code something like this

```python
async def stream_low():
    redis = await aioredis.create_redis('redis://localhost')

    while True:
        messages = await redis.xread(streams=["mystream"])
        for message in messages:
            print(message)
```
using redis stream in this way can be misleading because every call to `XREAD` will give you the messages sent after the `XREAD` call (because by default latest_ids is `$`) that means that during the time you are processing the messages if you don't track the last id, you will lose them (because the next call will use `give me the latest again`).

my changes add:
- support for async for
- support for encoding 
- keep track of the latest_id processed for each stream (future calls to `XREAD` will use the correct latest_ids)

```python

async def stream_async_for():
    redis = await aioredis.create_redis('redis://localhost')
    streams = redis.streams.consumer(["mystream"], encoding='utf-8')

    async for message in streams:
        print(message)
        print(streams.last_ids_for_stream)
```


with groups you would have to do something like this:

```python
async def stream_group_low():
    redis = await aioredis.create_redis('redis://localhost')
    latest_ids = ['0']
    while True:
        messages = await redis.xread_group(group_name="mygroup", consumer_name="Alice", streams=["mystream"], latest_ids=latest_ids)
        if messages == []:
            latest_ids = ['>']
        for message in messages:
            print(message)
            await redis.xack("mystream", "mygroup", message[1])
```

`latest_ids = ['0'] ` means send me the pending messages, while `latest_ids = ['>']` starts from the last you didn't send me.

my changes add:
- support for async for
- support for encoding 
- keep track of the latest_id processed for the stream
- if you want to check for pending messages you can use `check_pending`
- make acknowledging messages _nicer_

this is an example

```python
async def stream_async_for_group():
    redis = await aioredis.create_redis('redis://localhost')
    streams = redis.streams.consumer_with_group(["mystream"], group_name="mygroup", consumer_name="Alice", encoding='utf-8')

    async for message in streams:
        print(message)
        await streams.ack_message(message[1])
        print(streams.last_ids_for_stream)
```

seandstewart:
@barrachri -

Thanks so much for this contribution and apologies for the long wait. Maintenance has now resumed for this project!

I love the look of it and fully support adding a more intuitive interface for streaming. Would you be willing to rebase your changes on master? Let's get this utility merged in!

aioredis/streams_utils.py
```diff
@@ -0,0 +1,174 @@
...
+            streams=self._streams,
+            count=self._count,
+            latest_ids=latest_ids,
+            encoding=self._encoding
+        )
+
+        self._check_pending = False if len(messages) == 0 else True
+
+        logger.info("Received %d messages..." % len(messages))
+        for message in messages:
```
seandstewart, line 139: Maybe `await asyncio.gather(*(self._queue.put(m) for m in messages))`

aioredis/streams_utils.py
```diff
@@ -0,0 +1,174 @@
...
+        messages = await self._redis.xread(
+            streams=streams,
+            count=self._count,
+            latest_ids=latest_ids,
+            encoding=self._encoding
+        )
+
+        logger.info("Received %d messages..." % len(messages))
+
+        for message in messages:
```
seandstewart, line 154: Same as above ^^

aioredis/streams_utils.py
```diff
@@ -0,0 +1,174 @@
...
+            await self._queue.put(message)
+
+    async def get(self):
+        """Coroutine that waits for and returns a message.
+
+        :raises aioredis.ChannelClosedError: If channel is unsubscribed
+            and has no messages.
+        """
+        while self._queue.empty():
```
seandstewart, line 163: I like the queue pattern here, greatly simplifies the work of waiting.

aioredis/commands/streams.py
```diff
@@ -3,13 +3,17 @@
 from aioredis.util import wait_convert, wait_make_dict, wait_ok
 
 
-def fields_to_dict(fields, type_=OrderedDict):
+def fields_to_dict(fields, type_=OrderedDict, encoding=None):
```
seandstewart, line 6: Note to myself, would be good to move away from OrderedDict - that's completely unnecessary in Python 3.

aioredis/commands/streams.py
```diff
@@ -63,10 +67,16 @@ def parse_messages_by_stream(messages_by_stream):
     """
     if messages_by_stream is None:
         return []
-
     parsed = []
+
     for stream, messages in messages_by_stream:
```
seandstewart, line 72: I think I'd like to have two branches here - one that handles decoding and one that does not, so that we don't have to re-evaluate the existence of an encoding parameter for every message returned.

experiment.py
```diff
@@ -0,0 +1,79 @@
+import asyncio
```
seandstewart, line 1: I think we can add this to the `examples` folder, it's very useful to see the utility in action!

seandstewart (changes requested):
Thanks again for the submission, I'm really loving the look of this. Just a few comments to resolve and then I think we can ship it 🚢

barrachri:
@seandstewart sure 👍

seandstewart:
Hey @barrachri - I hate to do this, but I've been thinking over this PR and while I think the utility is very useful, I don't think it's in the spirit of aioredis, which is meant to simply be a thin client for communicating to redis. As such, I think our focus should be on providing a good API for executing Redis commands and managing Redis connections, rather than providing a runtime framework for Redis features.

I'm going to close this for now - apologies for the reversal on this one.

I'd love to have your help continuing maintenance of this library moving forward, however! Please consider taking a stab at any of our open issues.

PR #475, opened 2018-10-10
backport to python 3.5.1, without hiredis

giovanni-angeli:
in order to install this lib on an embedded system with python 3.5.1 and without hiredis, I had to make those patches.


popravich:
There is a little bit more than just a `create_future` (e864842fdb52feaa49718747648dc3cf758b54d4).
I don't really want to bring those wrappers back to `master` but I want to help you to use aioredis on embeded device.
Currently I see two major issues with this backport:
* running tests for Python 3.5.1 — travis-ci now has Python 3.5.6, so we would need some special setup to run tests.
* the 3.5.1 support itself — this would mean that we won't be able to use new language features
to implement library features in easy way.

I can imagine this backport living in a separate branch and being synced from master. However this is just a quick thought and I need to think it over because in this case the support (features backporting and releases) can become really hard.

giovanni-angeli:
Thank you Alexey for your answer. 
In effect, I understand that remaining constrained to the support of 3.5.1 would prevent to use new language features and this is too much.
A separate fork may be a good way to go. I can imagine that It will survive for a shorter period than the main project, i.e. soon or late even peculiar/embedded environments will be upgraded. 
Anyway, with those patches I can do what I need in the short/medium term, So take all your time to think it over. 
Thanks a lot for your time, Giovanni.

setup.py
```diff
@@ -6,7 +6,7 @@
 
 
 install_requires = ['async-timeout']
-if platform.python_implementation() == 'CPython':
+if platform.python_implementation() == 'CPython' and not os.environ.get("WITHOUT_HIREDIS"):
```
blueyed: Maybe "hiredis" should be moved into an `extras_require`?
(I assume this is not due to the Python version, but because hiredis should be skipped/is not available in general)

seandstewart:
@giovanni-angeli -

Thank you for your submission - however we have decided to deprecate support for Python3.5 moving forward.

PR #516, opened 2019-01-18
tests: upgrade pytest to 4.1.1

blueyed:
Uses stopgap measure for removed pytest_namespace hook.

Closes https://github.com/aio-libs/aioredis/pull/513.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -609,6 +609,11 @@ def pytest_configure(config):
                 "Can not import uvloop, make sure it is installed")
         asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
+    # Stopgap for removed pytest_namespace hook.
+    pytest.assert_almost_equal = assert_almost_equal
+    pytest.logs = logs
+    pytest.redis_version = redis_version
+
 
 def logs(logger, level=None):
     """Catches logs for given logger and level.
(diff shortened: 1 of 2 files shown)


webknjaz:
Restarted that job

blueyed:
Another one now:
```
_____________________ test_pool_idle_close[redis_v2.6.17] ______________________
[gw0] linux -- Python 3.7.1 /home/travis/virtualenv/python3.7.1/bin/python
create_pool = <function create_pool.<locals>.f at 0x7f1b811a6488>
start_server = <function start_server.<locals>.maker at 0x7f1b811a6378>
loop = <_UnixSelectorEventLoop running=False closed=False debug=False>
    @pytest.mark.run_loop
    async def test_pool_idle_close(create_pool, start_server, loop):
        server = start_server('idle')
        conn = await create_pool(server.tcp_address, minsize=2, loop=loop)
        ok = await conn.execute("config", "set", "timeout", 1)
        assert ok == b'OK'
    
        await asyncio.sleep(2, loop=loop)
    
        # On CI this test fails from time to time.
        # It is possible to pick 'unclosed' connection and send command,
        # however on the same loop iteration it gets closed and exception is raised
>       assert (await conn.execute('ping')) == b'PONG'
E       aioredis.errors.ConnectionClosedError: Reader at end of file
tests/pool_test.py:535: ConnectionClosedError
```

PR #517, opened 2019-01-19
Travis: use Ubuntu Xenial

blueyed:


.travis.yml
```diff
@@ -39,6 +35,7 @@ matrix:
   allow_failures:
   - env: UVLOOP="y"
   - python: "pypy3.5-5.8.0"
```
webknjaz: This could also be updated to 6.0.0

blueyed: I wasn't sure which versions are available etc.
Let's do it in another PR then?

webknjaz: Sure. It's poorly advertised which PyPy versions are available under Travis CI...
Still, this PR seems to fail with current changes

webknjaz: Oh, it was flaky

PR #522, opened 2019-01-23
Update dependencies

popravich:


setup.cfg
```diff
@@ -6,6 +6,10 @@ markers =
     run_loop: Mark coroutine to be run with asyncio loop.
     redis_version(*version, reason): Mark test expecting minimum Redis version
     skip(reason): Skip test
+python_files =
+    test_*.py
+    *_test.py
+    _testutils.py
```
blueyed: Hmm.. `_testutils.py` does not contain tests, does it?

PR #526, opened 2019-01-23
Update .travis.yml config + py-parser fix

popravich:


.travis.yml
```diff
@@ -5,50 +5,55 @@ addons:
...
+    name: flake
+    python: "3.6"
+    install:
+    - pip install -r tests/requirements.txt
     script:
-    - make spelling
-  - python: "pypy3.5-5.8.0"
-    env: PYPY_RELEASE="pypy3-v6.0.0-linux64"
-  - python: "3.6"
+    - make flake
```
blueyed: spelling and flake could also be run in a single job.

.travis.yml
```diff
@@ -5,50 +5,55 @@ addons:
...
     packages:
     - socat
-    - enchant
-    - aspell
-    - aspell-en
 
 env:
   global:
-  - REDIS_TAGS="2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1" INSTALL_DIR=$HOME/redis
+  - REDIS_TAGS="2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1" INSTALL_DIR=$HOME/redis TEST_ARGS="-n 4"
```
blueyed: Better `-n auto`?

aioredis/parser.py
```diff
@@ -63,9 +63,9 @@ def waitany(self):
         yield from self.waitsome(len(self.buf) + 1)
 
     def readone(self):
-        if not self.buf[self.pos:1]:
+        if not self.buf[self.pos:self.pos + 1]:
             yield from self.waitany()
-        val = self.buf[self.pos:1]
+        val = self.buf[self.pos:self.pos + 1]
```
blueyed: Can/should this have a test?

.travis.yml
popravich: I don't want to mess all tasks into single build step as it was before.
It is much easier to understand what has failed from travis dashboard than digging into
job's log.

aioredis/parser.py
popravich: Yes, it would be great.

.travis.yml
popravich: Probably... 
Once I run tests with `-n auto` and it started 32 processes, now when I was adding this change it run with only 2 processes...

.travis.yml
popravich: Also, will try to move examples into separate stage

webknjaz: Wow! At least someone supports me on non-messy jobs...

.travis.yml
webknjaz: AFAIR pytest-xdist is intelligent enough to identify Travis CI and set this to 2 processes. If it can access the corresponding env vars, of course...

webknjaz: Oh, and there's `PYTEST_ADDOPTS` special var for this purpose...

P.S. I like mapping-style var definition more, it's better readable.

.travis.yml
```diff
@@ -5,50 +5,55 @@ addons:
...
+- lint
+- test
+
+jobs:
   include:
-  - python: "3.6"
-    env: DOCS_SPELL_CHECK="y"
+  - stage: lint
+    name: documentation spell check
+    python: "3.6"
```
webknjaz: Let's do 3.7

.travis.yml
```diff
@@ -5,50 +5,55 @@ addons:
...
+    install:
+    - pip install -r tests/requirements.txt
     script:
-    - make spelling
-  - python: "pypy3.5-5.8.0"
-    env: PYPY_RELEASE="pypy3-v6.0.0-linux64"
-  - python: "3.6"
+    - make flake
+  - <<: *FLAKE
```
webknjaz: I think, one flake job is enough. The linter does the same thing regardless of runtime..

.travis.yml
blueyed: @webknjaz 
I also like that better, but there is also quite some overhead sometimes that could be avoided.
It is much better with CircleCI in general, where you have slim/custom Docker base images, and not a bunch of requirements need to be installed.

.travis.yml
```diff
@@ -5,63 +5,85 @@ addons:
...
     packages:
     - socat
-    - enchant
-    - aspell
-    - aspell-en
 
 env:
-  global:
-  - REDIS_TAGS="2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1" INSTALL_DIR=$HOME/redis
+- &ENV
```
webknjaz: This is a wrong syntax for global env. instead use:
```suggestion
  global:
    REDIS_TAGS: >-
      2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1
    INSTALL_DIR: >-
      $HOME/redis
    PYTEST_ADDOPTS: >-
      -n auto
```

.travis.yml
```diff
@@ -7,8 +7,10 @@ addons:
     - socat
 
 env:
-  global:
-  - REDIS_TAGS="2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1" INSTALL_DIR=$HOME/redis TEST_ARGS="-n 4"
+- &ENV
+  REDIS_TAGS: "2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1"
+  INSTALL_DIR: $HOME/redis
+  PYTEST_ADDOPTS: '"-n auto"'
```
webknjaz: a snippet above introduces a better readable version of wrapping quotes

.travis.yml
```diff
@@ -38,35 +51,39 @@ jobs:
...
+    - make examples
+  - <<: *EXAMPLES
     python: "3.7"
 
 install:
 - make -j ci-build-redis
 - |
     if [ "$UVLOOP" = "y" ]; then
-      export TEST_ARGS="$TEST_ARGS --uvloop"
+      export PYTEST_ADDOPTS="$PYTEST_ADDOPTS --uvloop"
```
webknjaz: by the way, there's also `pytest.ini` where you can set up additional options. if you need "xdist" enabled everywhere it'd  work out of the box

.travis.yml
```diff
@@ -7,7 +7,13 @@ addons:
...
 env:
-- &ENV
+  global:
+    REDIS_TAGS: >-
+      2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1
+    INSTALL_DIR: >-
+      $HOME/redis
+    PYTEST_ADDOPTS: >-
+      '-n auto'
   REDIS_TAGS: "2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1"
```
webknjaz: don't forget to remove these lines now

.travis.yml
```diff
@@ -5,63 +5,86 @@ addons:
...
     - socat
-    - enchant
-    - aspell
-    - aspell-en
 
 env:
   global:
-  - REDIS_TAGS="2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1" INSTALL_DIR=$HOME/redis
+    REDIS_TAGS: >-
+      2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1
```
webknjaz: Oh, I've made a mistake, it should be:
```suggestion
      '2.6.17 2.8.22 3.0.7 3.2.8 4.0.11 5.0.1'
```
(because of spaces)

blueyed:
Nice caching enhancements.
What about renaming `INSTALL_DIR` to `REDIS_INSTALL_BASEDIR` for clarity?

PR #528, opened 2019-01-23
tests: pool_or_redis: remove version check

blueyed:
It does not make sense to handle older versions of ourselves.


popravich:
It is there on purpose. There is no reason for this test if not compared with older version of library.
These are integration tests, they are needed to ensure that behavior between v0.3 and current v1
is either identical or mark the difference (expected failure with `blpop` and pool size equal 1).
This test can be (and is) cherry-picked into v0.3 branch to check compatibility.

PR #546, opened 2019-02-04
Add example for integration with aiohttp

gyermolenko:


seandstewart:
@gyermolenko -

Thanks for supplying this example! Would you be willing to update the example and documentation for latest master?

PR #592, opened 2019-05-07
In tests accept connection string from command line

gyermolenko:
This PR is created as a basis for discussion ( but can become a "Work in progress" :wink:  )

I wanted to test `aioredis` against different versions of `Redis` and couldn't find a way to do so. 
I mean, it is not easy to install one particular patch version of redis, let alone several. 

So I've added `--redis-url` cmdline option to use like so:
```text
$ pytest tests/string_commands_test.py::test_get -sq --redis-url="redis://localhost:6371"                                                                                                          
..
2 passed in 0.06 seconds
```
For this to work there should also be redis-server somewhere, e.g.:
```bash
$ docker run --rm -p 6371:6379 redis:5.0.3
```

Please let me know what you think.
And maybe there is a proper (simpler) way to test the lib against different versions of `Redis`, please let me know.

P.S. I have even drawn this doodle to understand what-goes-where, but no luck
image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -130,6 +130,19 @@ def redis(create_redis, server, loop):
 
 
 @pytest.fixture
+def redis_url(request):
+    return request.config.getoption("--redis-url")
+
+@pytest.fixture
+def redis_(create_redis, loop, redis_url):
+    """Returns Redis client instance."""
+    redis = loop.run_until_complete(
+        create_redis(redis_url, loop=loop))
+    loop.run_until_complete(redis.flushall())
+    return redis
+
+
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


popravich:
Hi, @gyermolenko 

> And maybe there is a proper (simpler) way to test the lib against different versions of Redis, please let me know.

There is a Makefile which know how to fetch, build and run tests with any redis version.
Try executing:
```bash
REDIS_TAGS="1.2.3 4.5.6" make ci-test -j  # -j for parallel build
```

seandstewart:
Hey @gyermolenko - thanks for the contribution! However, this change has been superseded by #891.

PR #594, opened 2019-05-10
Add demo of sharing pool from multiple coroutines

bsolomon1124:
## What do these changes do?

Add a small example in the README showing usage of a single `pool` from multiple coroutines.

In the current README, it is not obvious for someone new to async io that the connection should only be closed once from the top-level caller.


seandstewart:
@bsolomon1124 thank you for your contribution!

Issue #511, opened 2019-01-07
Blocking operations interfere with acquiring a pool connection


bbridges:
I'm currently having a problem where if a blocking operation is running in the background, and I'm needing to acquire a connection to use with `watch`/`unwatch` several times, eventually the connection waits until the blocking operation is done to use that connection instead of the other free connections. If it takes a while for the blocking operation to go through it prevents my transaction from happening.

Here's a minimal example that will cause this to happen:

```python
import asyncio

import aioredis


loop = asyncio.get_event_loop()


async def main():
    redis = await aioredis.create_redis_pool('redis://localhost', minsize=5,
                                                                  maxsize=10,
                                                                  loop=loop)

    # Run a background operation.
    redis.brpop('something')

    for i in range(0, 20):
        print(i)
        # Acquiring a connection and doing something.
        with await redis as r:
            await r.set('test', 1)

    redis.close()
    await redis.wait_closed()


loop.run_until_complete(main())
```

For me, the above runs the loop twice, waits until the blocking operation finishes, and then continues. Making the blocking task acquire its own connection manually will fix the problem, but it won't work without doing that in the above.

Edit: Similar behavior happens when not acquiring it manually, i.e. just running `redis.set('test', 1)` directly, but the loop runs 5 times instead before stopping.

adrianog:
The example you reported is similar to what I see in the documentation (https://goo.gl/7d9QZS), except:
1) The redis variable is used outside of the async coroutine (shouldn't matter)
2) There is no brpop example in that documentation (unfortunate as that is a clear blocking case).

In your case the "exclusive" mode should be applied to the brpop call, so:
```
# Exclusive mode
with await redis as r:
  await redis.brpop('something')

 for i in range(0, 20):
        print(i)
        # Shared mode
        await redis.set('test', 1)
```

bbridges:
Yeah I have to use

adrianog:
I'd call it a known limitation with an effective workaround.

PS: I think a lot of people come to aioredis with that blocking use case in mind, and I wish more prominence was given to this issue - why isn't brpop in the docs given that that is a glaring blocking example .

pbecotte:
Thanks, just came here to ask what I was doing wrong, but this solves what I needed.

PR #655, opened 2019-10-28
add example of using blocking commands

pbecotte:
Add an example of using blocking commands to the documentation.

#511

### DIFF
--- a/examples/blocking.py
+++ b/examples/blocking.py
@@ -1,0 +1,31 @@
+import asyncio
+
+import aioredis
+
+
+async def blocking_commands():
+    # Redis client bound to pool of connections (auto-reconnecting).
+    redis = await aioredis.create_redis_pool(
+        'redis://localhost')
+
+    async def get_message():
+        # Redis blocking commands are NOT naturally async. They block the
+        # connection that they are using. Therefore, if you exhaust a connection
+        # pool, blocking commands could wind up blocking the event loop. To
+        # avoid this unexpected behavior, always use explicit "acquire"
+        # syntax with blocking commands.
+        with await redis as r:
+            return await r.brpop("my-key")
+
+    future = asyncio.create_task(get_message())
+    await redis.lpush('my-key', 'value')
+    await future
+    print(future.result())
+
+    # gracefully closing underlying connection
+    redis.close()
+    await redis.wait_closed()
+
+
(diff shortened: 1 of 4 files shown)


examples/blocking.py
```diff
@@ -0,0 +1,31 @@
...
+    # Redis client bound to pool of connections (auto-reconnecting).
+    redis = await aioredis.create_redis_pool(
+        'redis://localhost')
+
+    async def get_message():
+        # Redis blocking commands are NOT naturally async. They block the
+        # connection that they are using. Therefore, if you exhaust a connection
+        # pool, blocking commands could wind up blocking the event loop. To
+        # avoid this unexpected behavior, always use explicit "acquire"
+        # syntax with blocking commands.
```
owtaylor: I think this explanation isn't quite right - even with the acquire syntax, you can still exhaust the connection pool if you have that many blocking commands running at once. What the acquire syntax prevents is the connection being returned to the pool until the blocking command completes - e.g. - if the lpush() command got queued on the same connection behind the brpop(), it would never run, because brpop() would never have anything to pop.

owtaylor: I should add that - in my not very informed opinion - this is very much automatable in the high-level commands interface returned by create_redis_pool() and that would be better than documenting how to work around it.

pbecotte: If you could suggest more accurate phrasing, I'd be happy to update this. However, considering that there was no working example of a blocking command in the docs, it seemed useful to add one instead of having to find a github issue with example code?

owtaylor: Something like:

"Redis blocking commands block the connection they are on until they complete. For this reason, the connection must not be returned to the connection pool until we've finished waiting on future created by brpop().  To achieve this, 'await redis' acquires a dedicated connection from the connection pool and creates a new Redis command object using it. This object is a context manager and the connection will be released back to the pool at the end of the with block."

I think it's great that you're contributing the example here!

pbecotte:
Updated the comment as suggested. Thanks for the clearer (and more accurate!) phrasing!

seandstewart:
> Updated the comment as suggested. Thanks for the clearer (and more accurate!) phrasing!

Looks good - want to add yourself to `CONTRIBUTORS.txt` and add an entry to the `CHANGES` directory? Then I think we can merge this in.

PR #701, opened 2020-02-18
Add string parameter support for min & max parameter of SortedSet commands.

### LINKED ISSUE #596: Why keyword arguments for SortedSet commands  should be bytes
Argument type for many SortedSet commands (for `min`/`max` kw) is enforced by isinstance check.
In one place it even has this comment:
```python
if not isinstance(max, bytes):  # FIXME     Why only bytes?
    raise TypeError("max argument must be bytes")
```

I think it is more convenient (for me as a user) to pass strings inst. of bytes.
Some of these commands are for "lexicographical" operations (so user mostly works with strings and not blobs of bytes)

Also strings would be more consistent with other commands.

Should I work on a PR for the change?
If there is a deep reason for that - could you please explain the logic behind it? 
Thanks.

undali:
This should fix #596.

### DIFF
--- a/aioredis/commands/sorted_set.py
+++ b/aioredis/commands/sorted_set.py
@@ -148,13 +148,19 @@ class SortedSetCommandsMixin:
         """Count the number of members in a sorted set between a given
         lexicographical range.
 
-        :raises TypeError: if min is not bytes
-        :raises TypeError: if max is not bytes
+        :raises TypeError: if min is not bytes or str
+        :raises TypeError: if max is not bytes or str
         """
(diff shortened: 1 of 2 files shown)


seandstewart:
@undali -

Thanks for your contribution! Would you be willing to resolve these merge conflicts?

Additionally, make sure to add an entry to the `CHANGES` directory and add yourself as a contributor!

seandstewart:
Hey @undali  - thanks for the contribution! However, this change has been superseded by #891.

PR #724, opened 2020-03-24
Fix some timeout tests for Python 3.8

aplanas:
## What do these changes do?

Some tests fails on Python 3.8 due to the side_effect being a lambda, that cannot be declared as async.

Also expectation for test_pool_idle_close do not match 3.8.2, so PBO_34638 was adjusted.

No, only for tests.

## Related issue number

N/A


seandstewart:
@aplanas -

Thanks so much for the submission! Maintenance of this library resumed this week.

These tests were actually resolved as a part of a separate effort #840 so this PR is no longer needed, but your contribution is still greatly valued!

If you'd like to get involved in the effort to resuscitate this library, I'd be happy to have you on board!

PR #743, opened 2020-05-11
Fixes aio-libs/aioredis#742, fixes aio-libs/aioredis#690

mohammadhzp:
## What do these changes do?
1- Starting with Redis 5, sentinel instance can accept a password to require an authentication too. aioredis can accept a password on its core but there was no way for a user to set a password for it

2- In case no sentinels found, sentinel pool closes properly now.

No.
Current PR is basically a few line edits which fix a bug and covers a gap. it doesn't break anything.


seandstewart:
@mohammadhzp would you be willing to fix any remaining test & lint issues? I'd like to see this merged in.

seandstewart:
Hey @mohammadhzp  - thanks for the contribution! However, this change has been superseded by #891.

PR #845, opened 2020-11-16
Fix syntax error introduced from code reformat

prdx23:
## What do these changes do?

I forked the repo today to submit a pr for another issue but it had syntax issues, which i think got introduced due to the recent format changes. I have added the deleted lines back after referencing the second to last commit and pytest also passes.

No

## Related issue number

None

## Checklist

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -11,6 +11,7 @@ Andrew Svetlov
 Andrew Chen Wang
 Anton Salii
 Anton Verinov
+Arsh <prdx23>
 Artem Mazur
 <cynecx>
 David Francos
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -168,9 +168,14 @@ class RedisConnection(AbcConnection):
             warnings.warn("The loop argument is deprecated", DeprecationWarning)
         if parser is None:
             parser = Reader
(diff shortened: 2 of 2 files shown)


Andrew-Chen-Wang:
I'd like to close this since we're slowly merging a lot of PRs right now, and majority PRs won't be compatible with Travis CI until our GitHub actions implementation. I think once we're at a good spot / enough changes, we can do a final lint.

prdx23:
Currently all tests pass on Travis CI except for one - `test_connection_idle_close` which tests idle timeout.  This test seems to fail arbitrarily on repeated runs of pytest which looked strange to me, so I had a look though the redis documentation and found this -

```
Timeouts are not to be considered very precise: Redis avoids to set timer events or to run O(N) algorithms in order to check idle clients, so the check is performed incrementally from time to time. This means that it is possible that while the timeout is set to 10 seconds, the client connection will be closed, for instance, after 12 seconds if many clients are connected at the same time.

```
Source: https://redis.io/topics/clients

I think we can increase the margin of error of that specific test to fix this issue. 

Other than that, let me know if you want this pr closed, I can use this branch locally until then to test my fixes and wait for you guys to get CI/CD set up.

Andrew-Chen-Wang (approved):
I think in retrospect, I take what I said back. I was just kinda alerted by the sudden additions of an assert and other declaration. So LGTM

PR #849, opened 2020-11-19
Change Z*LEX APIs to pass min and max straight through

### LINKED ISSUE #784: ZRANGEBYLEX does not handle min of "+" or max of "-"
The code for handling the special values `+` and `-` in zrangebylex only handles `-` for `min` and `+` for max, not vice versa. For example, the following code prints `[b'bar']` instead of an empty list:

```python
import asyncio
import aioredis

async def main():
    client = await aioredis.create_redis_pool('redis://localhost')
    await client.zadd('foo', 0.0, 'bar')
    result = await client.zrangebylex('foo', b'+', b'+')
    print(result)
    client.close()
    await client.wait_closed()

loop.run_until_complete(main())
loop.close()
```

This also affects `zrevrangebylex` and `zremrangebylex`.

Incidentally, the API design is problematic because it's impossible to pass strings like `(-` (meaning, strings lexicographically greater than `-`) into redis.

bmerry:
## What do these changes do?

This fixes #784 and eliminates the ambiguity where "-" could mean either
"from the beginning" or "from the literal string -". It is a breaking
API change.

Yes. Users will need to prefix the `min` and `max` value with `[` or `(`.

## Related issue number

#784 

## Checklist

### DIFF
--- a/aioredis/commands/sorted_set.py
+++ b/aioredis/commands/sorted_set.py
@@ -141,21 +141,10 @@ class SortedSetCommandsMixin:
         fut = self.execute(b"ZINTERSTORE", destkey, numkeys, *args)
         return fut
 
-    def zlexcount(self, key, min=b"-", max=b"+", include_min=True, include_max=True):
+    def zlexcount(self, key, min=b"-", max=b"+"):
         """Count the number of members in a sorted set between a given
         lexicographical range.
-
-        :raises TypeError: if min is not bytes
-        :raises TypeError: if max is not bytes
         """
-        if not isinstance(min, bytes):  # FIXME
-            raise TypeError("min argument must be bytes")
-        if not isinstance(max, bytes):  # FIXME     Why only bytes?
-            raise TypeError("max argument must be bytes")
-        if not min == b"-":
-            min = (b"[" if include_min else b"(") + min
-        if not max == b"+":
-            max = (b"[" if include_max else b"(") + max
         return self.execute(b"ZLEXCOUNT", key, min, max)
 
     def zrange(self, key, start=0, stop=-1, withscores=False, encoding=_NOTSET):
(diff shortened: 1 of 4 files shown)


bmerry:
Since this is a breaking API change I'm guessing it should also go into a migration document at some stage. Is there already one (other than the 0.3 -> 1.0 doc) or should I start one?

seandstewart:
> Since this is a breaking API change I'm guessing it should also go into a migration document at some stage. Is there already one (other than the 0.3 -> 1.0 doc) or should I start one?

That's a good call, there isn't currently a migration page. I'm sure this won't be the first breaking change we merge in. We're targeting a 2.0 release with all of these merges.

aioredis/commands/sorted_set.py
```diff
@@ -389,30 +352,17 @@ def zrevrangebylex(
         key,
         min=b"-",
         max=b"+",
-        include_min=True,
-        include_max=True,
         offset=None,
```
seandstewart, line 355: Since we're breaking the API anyway, what do you think about using the same parameter names as redis-py?

https://github.com/andymccurdy/redis-py/blob/2c9f41f46991f94f0626598600d1023d4e12f0bc/redis/client.py#L2803-L2817

seandstewart:
Just one question for you, but otherwise looks great! 

Thanks for taking the initiative and getting this fixed.

aioredis/commands/sorted_set.py
bmerry: Interesting question. It's not obvious what the right answer is, but with a bit of thought, my feeling is that we shouldn't change it, because SORT is in the same situation (`offset`/`count` in aioredis, `start`/`num` in redis-py). I don't really want to break SORT unless there is going to be an incompatible change anyway, and if we change Z*LEX but not SORT then aioredis isn't even internally consistent.

Incidentally the Redis command reference uses the terms "offset" and "count" (although that's just documentation, not protocol), so we'd just be trading consistency with one thing for consistency with another.

seandstewart (approved):

PR #891, opened 2021-01-25
Port redis-py impl and tests to aioredis.

seandstewart:
## What do these changes do?

This is a full port of [redis-py](https://github.com/andymccurdy/redis-py)'s client implementation into an asyncio-friendly implementation.

The approach is essentially a line-by-line re-implementation. The major caveats to this are:

1. Some connection code is simply incompatible and had to be implemented differently.
2. We use black on our code-base, which will create line-count discrepancies between the two sources.
3. I've added a first-pass at type annotations (a lot of work is needed to tighten it up if we want to add mypy to our CI, but it gets us much further along).

Beyond that, all _functional_ logic remains as close to the original, synchronous source as possible.

The motivation behind this is to ease the workload for future contributors and maintainers. By replicating redis-py's implementation, we greatly reduce our own workload, and are able to quickly port new features and bugfixes as they are added to redis-py, including tests.

## Are there changes in behavior for the user?

Yes.

## Related issue number

#822 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is `<Name> <Surname>`.
  * Please keep alphabetical order, the file is sorted by names.
- [x] Add a new news fragment into the [`CHANGES/`](../tree/master/CHANGES) folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example:
   `Fix issue with non-ascii contents in doctest text files.`

PR #910, opened 2021-02-26
KEEPTTL in set()

kirlf:
## What do these changes do?
keepttl in set()

``keepttl`` if True, retain the time to live associated with the key.
            (Available since Redis 6.0)

No, optional parameter only.

## Related issue number

No.


seandstewart:
Hey @kirlf - thanks for the contribution! However, this change has been superseded by #891.

PR #924, opened 2021-03-22
Fix async_timeout not awaited (Fixes #923)

Andrew-Chen-Wang:
## What do these changes do?

Async timeout should be properly awaited.

## Are there changes in behavior for the user?

This should fix a bug

## Related issue number

Fixes #923 

## Checklist

- [X] I think the code is well written
- [X] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [X] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is `<Name> <Surname>`.
  * Please keep alphabetical order, the file is sorted by names.
- [ ] Add a new news fragment into the [`CHANGES/`](../tree/master/CHANGES) folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example:
   `Fix issue with non-ascii contents in doctest text files.`

PR #925, opened 2021-03-22
Use pre-spawned disconnect coro in Connection.__del__

### LINKED ISSUE #923: RuntimeWarning: coroutine 'Connection.disconnect' was never awaited
When running tests that need mocking an `aioredis` connection, the following Runtime Warning gets triggered each time, and is mumbled together with some random code, which makes me think it looks like a fairly low-level bug:

```
tests/foo/test_foo_bar.py ...                                    [100%]
/usr/local/lib/python3.9/site-packages/pkg_resources/_vendor/packaging/version.py:490:
RuntimeWarning: coroutine 'Connection.disconnect' was never awaited
  reversed(list(itertools.dropwhile(lambda x: x == 0, reversed(release))))
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

I'm not sure where that comes from but not from my own code, as I never mention a `disconnect` method anywhere.

```python
import pytest
import aioredis

from foo import api


@pytest.fixture
async def aioredis_conn():
    conn = await aioredis.Redis(
        decode_responses=True,
        single_connection_client=True,
        **config.redis_test_credentials
    )
    yield conn
    await conn.flushdb()


@pytest.fixture
def mock_redis_conn(mocker, redis_conn, aioredis_conn):
    mocker.patch("foo.api.aioredis_conn", new=aioredis_conn)

def test_foobar(mock_redis_conn):
    do_something()
```

Am I doing this approximately right? Is there any specific reason why this warning occurs?

seandstewart:
## What do these changes do?

This prevents a leaked coroutine when the Connection is garbage-collected.

No

## Related issue number

resolves #923 

## Checklist

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -677,7 +677,7 @@ class Connection:
 
     async def _connect(self):
         """Create a TCP socket connection"""
-        with async_timeout.timeout(self.socket_connect_timeout):
+        async with async_timeout.timeout(self.socket_connect_timeout):
             reader, writer = await asyncio.open_connection(
                 host=self.host, port=self.port, ssl=self.ssl_context, loop=self._loop
             )
@@ -844,7 +844,7 @@ class Connection:
     async def read_response(self):
         """Read the response from a previously sent command"""
         try:
-            with async_timeout.timeout(self.socket_timeout):
+            async with async_timeout.timeout(self.socket_timeout):
                 response = await self._parser.read_response()
         except asyncio.TimeoutError:
             await self.disconnect()
@@ -1071,7 +1071,7 @@ class UnixDomainSocketConnection(Connection):  # lgtm [py/missing-call-to-init]
         return pieces
 
     async def _connect(self):
-        with async_timeout.timeout(self._connect_timeout):
+        async with async_timeout.timeout(self._connect_timeout):
             reader, writer = await asyncio.open_unix_connection(path=self.path)
         self._reader = reader
         self._writer = writer
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -632,7 +632,7 @@ class Connection:
                 if loop.is_running():
                     loop.create_task(coro)
                 else:
-                    loop.run_until_complete(self.disconnect())
+                    loop.run_until_complete(coro)
         except Exception:
             pass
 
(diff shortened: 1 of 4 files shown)


Andrew-Chen-Wang (approved):
Hm Idk if it's leaked per se in memory, but this change is definitely needed.

Andrew-Chen-Wang:
Mostly because I wasn't aware that storing the coroutine as an object would cause a leak. It also doesn't resolve #923 since it's still showing up in the unit tests... I'm going to update #924 because there are a couple more async_timeouts that still use sync `with`

Andrew-Chen-Wang:
Oh it's already merged haha there's a couple more. Do you wanna put that in a separate PR or straight merge?

seandstewart:
> Oh it's already merged haha there's a couple more. Do you wanna put that in a separate PR, this PR, or straight commit to master?

Haha whoops, yeah, I'll add them here

seandstewart:
@Andrew-Chen-Wang 

I've been testing locally and I found why the warning is being raised in tests after fixing up everything. It seems the issue is with the non-standard order of events when setting up and tearing the event loop in our tests. When I run the following script (Python 3.8) I see no more warnings:

```python
import asyncio
import aioredis


async def main():
    redis = aioredis.Redis()
    for _ in range(10):
        c = await redis.connection_pool.get_connection("_")
        await c.connect()


if __name__ == '__main__':
    asyncio.run(main(), debug=True)
```

Andrew-Chen-Wang:
> non-standard order of events when setting up and tearing the event loop in our tests

Huh based on the changed code, that's quite confusing. Besides `cluster` becoming an async fixture, I don't quite understand how the teardown methodology was incorrect? Or is it not, but instead the change of the teardown portion was more for being more concise?

Overall, nice work!

seandstewart:
> > non-standard order of events when setting up and tearing the event loop in our tests
> 
> Huh based on the changed code, that's quite confusing. Besides `cluster` becoming an async fixture, I don't quite understand how the teardown methodology was incorrect? Or is it not, but instead the change of the teardown portion was more for being more concise?
> 
> Overall, nice work!

Basically, we were creating a task in the loop but the loop was closing before the task was properly run. In a proper production environment this wouldn’t happen because part of the recommended pattern (and part of `asyncio.run` Is to cancel all outstanding tasks during shutdown.

Making the fixture async and a yielding fixture ensures the correct order of operations on tear down.

Andrew-Chen-Wang:
gotcha thanks!

PR #946, opened 2021-04-14
fix ssl argument in _connect method

stanislau-arkhipenka:
asyncio.open_connection requires SSLContext as ssl argument, but RedisSSLContext given. RedisSSLContext has .get() method, which produces SSLContext object. Using this method to provide right argument


stanislau-arkhipenka:
Hey,
Why the checks failed now?

Andrew-Chen-Wang (approved):
LGTM (we have some weird Timeline protection CI that isn't removed from this repo; nothing I can do about it).

PR #953, opened 2021-04-22
Allow ssl connection to redis when use sentinel

stanislau-arkhipenka:
This diff allows to pass ssl=True inside **connection_kwargs for Sentinel class to enable ssl connection to redis
Besides enabling functionality itself this change also makes experience with Sentinel() more consistent with Redis()

Now SentinelManagedConnection enherits from SSLConnection, instead of Connection. 
This allows to populate ssl related properties from SSLConnection class. 
if ssl=True SentinelManagedConnection uses parents constructor (SSLConnection)
If ssl=False SentinelManagedConnection uses grandparents constructor (Connection class)

The only downside of this implementation, is in case ssl=False instance of SentinelManagedConnection has dangling ssl-cert related properties.This properties does not affect behavior in any considerable way.
Afaik there is no obvious way to overcome this issue.

Users now can pass SSL related attributes to **connection_kwargs in Sentinel class (or alternativelly to master_for or slave_for methods)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,14 +44,14 @@ Benchmarks can be found here:
 ## Contribute
 
 -   Issue Tracker: <https://github.com/aio-libs/aioredis/issues>
--   Google Group: <https://groups.google.com/forum/>\#!forum/aio-libs
+-   Google Group: <https://groups.google.com/g/aio-libs>
 -   Gitter: <https://gitter.im/aio-libs/Lobby>
 -   Source Code: <https://github.com/aio-libs/aioredis>
--   Contributor's guide: devel
+-   Contributor's guide: [devel](docs/devel.md)
 
 Feel free to file an issue or make pull request if you find any bugs or
 have some suggestions for library improvement.
 
 ## License
 
-The aioredis is offered under a [MIT License](license.md).
+The aioredis is offered under a [MIT License](LICENSE).
--- a/aioredis/sentinel.py
+++ b/aioredis/sentinel.py
@@ -3,7 +3,7 @@ import weakref
 from typing import AsyncIterator, Iterable, Mapping, Sequence, Tuple, Type
 
 from aioredis.client import Redis
-from aioredis.connection import Connection, ConnectionPool, EncodableT, SSLConnection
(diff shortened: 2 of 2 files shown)


stanislau-arkhipenka:
@Andrew-Chen-Wang any thoughts regarding this request?

Andrew-Chen-Wang:
@stanislau-arkhipenka apologies for being super late. I'll look into this PR after finals, but take note of the PR in redis-py first: https://github.com/andymccurdy/redis-py/issues/1306 I haven't take too much of a look through there, but it seems like they're doing some inheritance. Try and follow along and see if there are any diffs.

Andrew-Chen-Wang (changes requested):
@stanislau-arkhipenka Do you mind replicating what's been done here instead of using an "ssl" option to be more close with Redis implementation: https://github.com/andymccurdy/redis-py/pull/1419

It's not been reviewed by @andymccurdy yet, but the implementation based on the original issue seems to be The Way: https://github.com/andymccurdy/redis-py/issues/1306

stanislau-arkhipenka:
> 
> 
> @stanislau-arkhipenka Do you mind replicating what's been done here instead of using an "ssl" option to be more close with redis-py implementation: [andymccurdy/redis-py#1419](https://github.com/andymccurdy/redis-py/pull/1419)
> 
> That PR's not been reviewed by @andymccurdy yet, but the implementation and comments made by him and others (and others' implementation like that at Celery) based on the original issue seems to be The Way: [andymccurdy/redis-py#1306](https://github.com/andymccurdy/redis-py/issues/1306)

Hey, actually [originally ](https://github.com/aio-libs/aioredis-py/pull/953/commits/7e82bee99520cfdfa2bf7f7926f6502f26c647ad) i did pretty similar diff, but LGTM was not very happy because of [diamand problem](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem). MRO seems solves this problem, but it'll be definitely harder to read the code.
Revert?

Andrew-Chen-Wang (approved):
Thanks for the PR and fixing up the LGTM!

PR #956, opened 2021-04-28
Fix codecov paths

Andrew-Chen-Wang:
Codecov does not pick up on all files, e.g. aioredis/connection.py located at this commit in codecov: https://codecov.io/gh/aio-libs/aioredis-py/tree/b9694797d452bb8d4ca26d2dc14ce73068575130. Resolve via comment in codecov forum: correct fix: https://community.codecov.io/t/codecov-report-missing-files/1928/9 original fix: ~~https://community.codecov.io/t/missing-files-in-codecov-but-not-in-report/825/9~~

PR #957, opened 2021-04-28
Fix codecov folder structure

Andrew-Chen-Wang:
Supersedes #956 to resolve the comments about how you couldn't actually view the files...

Final solution comes from: https://github.com/thomasrockhu/aioredis-py/pull/1

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,0 @@
-codecov:
-  disable_default_path_fixes: yes
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,2 @@
-fixes:
-  - "::aioredis/"
-  - "::tests/"
+codecov:
+  disable_default_path_fixes: yes
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+fixes:
+  - "home/runner/work/aioredis-py/aioredis-py/::"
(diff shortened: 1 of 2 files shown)


Andrew-Chen-Wang:
nope

Andrew-Chen-Wang:
arg...

AraHaan:
Just so you know there was a security notice on codecov as well. As such I migrated from codecov to codacy, and it integrates well with github actions too.

Despite my project being .NET, it supports python, C, C++, ext. and as such I think it's a better alternative to codecov that is more secured.

This also means all the aio-libs repositories might need to invalidate every authentication token that may have been exposed by the hacked bash runners that would send the things in $env over to their external servers.

Andrew-Chen-Wang:
@AraHaan thanks for the comment. You can suggest codacy over at https://aio-libs.discourse.group/ I'm not head of this repo -- just a broke college student that wants to fix this damn bug... I think the comment about being "more secure" isn't really necessary in the context of open source projects. Hacks happen all the time, and we don't pay money for open source libs so I think we're fine. But thanks for the comment!

Andrew-Chen-Wang:
Solution fixed: https://github.com/thomasrockhu/aioredis-py/pull/1 Thanks to tom on codecov! Forum explanation: https://community.codecov.io/t/cant-view-files-on-codecov/2833/2

seandstewart (approved):

PR #959, opened 2021-04-29
Fix socket timeout

stanislau-arkhipenka:
This diff fixes a bug related to socket_timeout parameter. 
Without this fix attempt to set socket_timeout any value other than 0 results into following error:
`
  File "aioredis/connection.py", line 654, in connect
  File "aioredis/connection.py", line 700, in _connect
  File "/usr/local/lib/python3.8/asyncio/trsock.py", line 188, in settimeout
    raise ValueError(
ValueError: settimeout(): only 0 timeout is allowed on transport sockets
`


stanislau-arkhipenka:
Messed up with branches

PR #976, opened 2021-05-11
Avoid AttributeError from Pipeline.reset() with a ConnectionError.

egnor:
## What do these changes do?

Avoids an AttributeError from being thrown if in the process of attempting UNWATCH in Pipeline.reset(), the connection is closed and Pipeline.connection is set to None.

No changes.

## Related issue number

No known issue (though I could open one).

## Checklist

### DIFF
--- a/aioredis/client.py
+++ b/aioredis/client.py
@@ -4291,7 +4291,8 @@ class Pipeline(Redis):  # lgtm [py/init-calls-subclass]
                 await self.connection.read_response()
             except ConnectionError:
                 # disconnect will also remove any previous WATCHes
-                await self.connection.disconnect()
+                if self.connection:
+                    await self.connection.disconnect()
         # clean up the other instance attributes
         self.watching = False
         self.explicit_transaction = False


egnor:
Note, I don't know how this situation is happening (there's at least one other error in play), but this seems like a good defensive check to me...

Andrew-Chen-Wang:
It could be due to https://github.com/aio-libs/aioredis-py/blob/965b4102ec7530179c92e0bc6e1462a78c55cf2f/aioredis/client.py#L4343-L4356 because the only other times self.connection could be set to None is if either of these two are having an error: https://github.com/aio-libs/aioredis-py/blob/965b4102ec7530179c92e0bc6e1462a78c55cf2f/aioredis/client.py#L4290-L4291

since we do an if check beforehand: https://github.com/aio-libs/aioredis-py/blob/965b4102ec7530179c92e0bc6e1462a78c55cf2f/aioredis/client.py#L4286

Andrew-Chen-Wang:
I guess that's more of a code design issue. I'll leave this to seandstewart to decide a merge or not.

seandstewart (approved):

seandstewart:
This is a good thing to guard against in any case, since it could mask the issue

> Aha, I think I found why I kept exercising weird weird code paths -- I was running _two concurrent_ calls to get() on the same Pipeline. (Some code got migrated from being outside a pipeline to being inside one, and was doing some things in parallel...)
> 
> Of course, everything in a Pipeline shares a connection so you can't actually do concurrent operations, and everything gets all tangled up when you try. (Ideally this would be caught, but maybe that's hard or slow.)
> 
> So this is totally operator error -- but having that defensive check in there brought me a step closer to the real issue, so it was at least helpful in that respect :-). Feel free to accept or discard as you please.

This is a good thing to guard against I think.

egnor:
(Should I go ahead and merge this? I'm not sure what your process is.)

Issue #951, opened 2021-04-21
using socket_timeout option leads to  ValueError (AIORedis 2.0)


ivan-antonyuck:
When running this code:
```
key = 'test'
redis = aioredis.Redis(connection_pool=aioredis.ConnectionPool.from_url(redis_url, socket_timeout=1))
await redis.get('key')
```

i'm getting the following exception:
```
...
  val = await redis.get(key)
  File "/usr/local/lib/python3.8/site-packages/aioredis/client.py", line 1034, in execute_command
    conn = self.connection or await pool.get_connection(command_name, **options)
  File "/usr/local/lib/python3.8/site-packages/aioredis/connection.py", line 1361, in get_connection
    await connection.connect()
  File "/usr/local/lib/python3.8/site-packages/aioredis/connection.py", line 662, in connect
    raise ConnectionError(exc) from exc
aioredis.exceptions.ConnectionError: settimeout(): only 0 timeout is allowed on transport sockets
```

This might be because of double timeout handling - socket_timeout option used both in _connect (where the exception is generated) and in disconnect, read_response with async_timeout. Should those be two separate options, one for socket and another for async_timeout?

Andrew-Chen-Wang:
In the middle of finals, so just questioning in advance: are you using latest on master or 2.0.0a1?

ivan-antonyuck:
I've tested with both, and the trace i've included was in fact from lastest on master.

Just tested again with 2.0.0a1 from pypi, and it seems the same.

Here's the repo with test case: https://github.com/ivan-antonyuck/aioredis-timeout-test/tree/master

stacktrace for the pypi version:

```
/Users/antonyuck/.local/share/virtualenvs/aioredis_tests-k2Rjn665/bin/python /Users/antonyuck/work/projects/aioredis_tests/test.py
Traceback (most recent call last):
  File "/Users/antonyuck/.local/share/virtualenvs/aioredis_tests-k2Rjn665/lib/python3.8/site-packages/aioredis/connection.py", line 654, in connect
    await self._connect()
  File "/Users/antonyuck/.local/share/virtualenvs/aioredis_tests-k2Rjn665/lib/python3.8/site-packages/aioredis/connection.py", line 697, in _connect
    sock.settimeout(self.socket_timeout)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/trsock.py", line 188, in settimeout
    raise ValueError(
[...]

PR #986, opened 2021-05-17
Remove socket timeout setup

ivan-antonyuck:
setting timeout on asyncio socket leads to an Exception being thrown

## What do these changes do?

This changes remove settimeout call as suggested by @Andrew-Chen-Wang 

No

## Related issue number

https://github.com/aio-libs/aioredis-py/issues/951#issuecomment-827322087

### DIFF
--- a/CHANGES/951.bugfix
+++ b/CHANGES/951.bugfix
@@ -1,0 +1,1 @@
+removed socket timeout setup
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -29,6 +29,7 @@ Hugo <hugovk>
 Ihor Gorobets
 Ihor Liubymov
 Ilya Samartsev
+Ivan Antonyuck
 James Hilliard
 Jan Špaček
 Jeff Moser
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -695,9 +695,6 @@ class Connection:
                     sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
                     for k, v in self.socket_keepalive_options.items():
                         sock.setsockopt(socket.SOL_TCP, k, v)
-                # set the socket_timeout now that we're connected
-                if self.socket_timeout is not None:
-                    sock.settimeout(self.socket_timeout)
 
             except (OSError, TypeError):
                 # `socket_keepalive_options` might contain invalid options


Andrew-Chen-Wang:
@ivan-antonyuck thanks for checking in. So this persists even on current master branch? Did you make sure to install from master first? Just double checking in case I'll need to do a regression from #960

Andrew-Chen-Wang:
I certainly wish a test with like a mock would be here but not sure if it's possible

ivan-antonyuck:
@Andrew-Chen-Wang i've tested with the latest master. It seems that the exception in question is always risen if timeout is set on the asyncio socket, that's why i've removed the timeout setting.

Here's the code for the method in question: https://github.com/python/cpython/blob/834498e178684a7e2da49b4efe1adea33e0026b0/Lib/asyncio/trsock.py#L185

fjbsantiago:
Are you guys counting on packaging and publishing these fixes to PyPi?

Andrew-Chen-Wang:
@fjbsantiago yes including this one before 2.0.0. I just need to double check (and maybe @abrookins can double check) that

1. this PR actually fixes the problem (which it seems it does), 
2. we don't need a regression from the previous PR that attempted to fix this (i.e. #960 
3. not only does this fix the problem but there is also a solution that is already in the code or needs to be implemented

abrookins:
@Andrew-Chen-Wang Been backed up this week! I'll read through this in the morning.

ivan-antonyuck:
Hello @abrookins, @fjbsantiago 

My concern and the cause for this PR is this line
https://github.com/aio-libs/aioredis-py/blob/fa5ef8ad281c83e91d0c10881411a17cd369a6a9/aioredis/connection.py#L700

This invokes this method in asyncio transport socket
https://github.com/python/cpython/blob/834498e178684a7e2da49b4efe1adea33e0026b0/Lib/asyncio/trsock.py#L185

which always throws an exception if timeout is not zero.

This renders the timeout feature unusable, which is inconvenient for production use.

abrookins:
Ok, I dug into this a bit. We can't set a timeout on the socket, but we're already handling the timeout correctly here: https://github.com/aio-libs/aioredis-py/blame/master/aioredis/connection.py#L820-L824

Ivan's change removes our erroneous call to `sockettimeout` without breaking timeouts. You can generate a timeout using his code pointed at a URL that drops connections:

```
# -*- coding: UTF-8 -*-
__author__ = u'antonyuck'
import aioredis
import asyncio
async def test():
    redis = aioredis.Redis(connection_pool=aioredis.ConnectionPool.from_url("redis://www.google.com:81", socket_timeout=5))
    await redis.get('key')

asyncio.get_event_loop().run_until_complete(test())
```

If you run it:

```
[...]

  File "/Users/andrewbrookins/src/aioredis-timeout-test/test.py", line 9, in <module>
    asyncio.get_event_loop().run_until_complete(test())
  File "/usr/local/Cellar/python@3.9/3.9.2_4/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/andrewbrookins/src/aioredis-timeout-test/test.py", line 7, in test
    await redis.get('key')
  File "/Users/andrewbrookins/src/aioredis-py/aioredis/client.py", line 1034, in execute_command
    conn = self.connection or await pool.get_connection(command_name, **options)
  File "/Users/andrewbrookins/src/aioredis-py/aioredis/connection.py", line 1359, in get_connection
    await connection.connect()
  File "/Users/andrewbrookins/src/aioredis-py/aioredis/connection.py", line 658, in connect
    raise TimeoutError("Timeout connecting to server")
aioredis.exceptions.TimeoutError: Timeout connecting to server
```

Andrew-Chen-Wang (approved):
Gotcha, thanks @ivan-antonyuck and @abrookins !!

PR #970, opened 2021-05-09
Add proper installation instructions in docs

Andrew-Chen-Wang:
## What do these changes do?

Shows proper installation instructions for the alpha version of aioredis which pypi doesn't default install.

## Are there changes in behavior for the user?

Yes

## Related issue number

#930 

Specifically https://github.com/aio-libs/aioredis-py/issues/930#issuecomment-835443521

## Checklist

- [X] I think the code is well written
- [ ] Unit tests for the changes exist
- [X] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is `<Name> <Surname>`.
  * Please keep alphabetical order, the file is sorted by names.
- [ ] Add a new news fragment into the [`CHANGES/`](../tree/master/CHANGES) folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example:
   `Fix issue with non-ascii contents in doctest text files.`

-----
[View rendered docs/index.md](https://github.com/aio-libs/aioredis-py/blob/docs-install-fix/docs/index.md)

PR #990, opened 2021-05-25
Update getting-started.md

RileyMShea:
The docs don't work for v1, which is currently what pip installs by default


Andrew-Chen-Wang:
Duplicate #970

Issue #1002, opened 2021-06-08
[2.0] Explicit loop parameters are deprecated


bmerry:
The new 2.0 implementation has a number of places where it either takes an explicit `loop` parameter, or passes one to an asyncio function. The need for this was largely removed in Python 3.5.3 (because `get_event_loop` was changed to return the event loop of the current task when called from inside a task), and many functions (such as [wait_for](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for) deprecated the parameter in 3.8 with removal for 3.10. Other functions (like [open_connection](https://docs.python.org/3/library/asyncio-stream.html#asyncio.open_connection)) didn't deprecate it, but made it optional and documented that it is not needed when called from a coroutine. I wouldn't be surprised if they are deprecated in future.

I would suggest:
(a) eliminating passing the parameter where it is deprecated e.g. wait_for
(b) eliminating passing the parameter where it is unnecessary
(c) eliminating the `loop` argument to the `Connection` constructor and the `_loop` attribute. The finalizer currently uses it, but I've argued in #930 that this is a bad idea.

Let me know if a PR is likely to be looked at.

Andrew-Chen-Wang:
@bmerry Do it! (asyncio is an evolving lib so yea go for it!)

Issue #930, opened 2021-03-28
Migration to aioredis 2.0


Andrew-Chen-Wang:
Hi all,

After #891, aioredis will be making a major jump to `aioredis==2.0.0`. The PR made aioredis an enormous async port of redis-py. For the migration docs, please refer to: https://github.com/aio-libs/aioredis-py/blob/master/docs/migration.md. To install:

```shell
pip install aioredis==2.0.0
```

<details><summary>Old pip install instructions for beta left for posterity</summary>

`pip install aioredis==2.0.0b1`

To install aioredis at a specific commit:

`$ pip install git+git://github.com/aio-libs/aioredis-py.git@ eb4c001a598c01d9a0e5f2a24065c486c823d1b0 `

To install at master:

`pip install git+git://github.com/aio-libs/aioredis-py.git@master`

</details>

We'd love to hear how well this new version of aioredis integrates with your current production code! @abrookins and I will be updating this specific GitHub issue with any updates that you should be aware of; any breakages can be reported as a separate GitHub issue or in this one, too.

Thanks everyone!

bmoscon:
I've been using this for a week or two (just running off a commit hash from that time period). It would be nice to be able to push my project up to pypi, but I can't because pypi wont let you use a direct dependency. Have you considered releasing a 2.0.0a preview version (or something like that)?

Andrew-Chen-Wang:
Yes we've considered it, but I don't have release permission only seandstewart, but @abrookins might? @abrookins I suggest we push out the alpha version. It won't show up as an upgrade to people but it at least makes the installation easier to do.

bmoscon:
You can see the list of those with access here, there are only 3 

https://pypi.org/project/aioredis/

bmoscon:
Hey @asvetlov - you're on the pypi account, any chance you could give access to the two Andrews so we can continue release work while the Sean is AFK? He made them admins on the repo, I doubt he'd be opposed to having them have pypi access

seandstewart:
@bmoscon - 

thanks for stepping in with your vote of confidence. I’ve been afk for the last ten days because my father passed away and I’m helping my mother with everything. I commented on the Maintainership ticket letting folks know. 

https://github.com/aio-libs/aioredis-py/issues/822#issuecomment-808465684

[...]

PR #1006, opened 2021-06-09
Remove unneeded `loop` arguments

bmerry:
## What do these changes do?

The new 2.0 implementation has a number of places where it either takes an explicit loop parameter, or passes one to an asyncio function. The need for this was largely removed in Python 3.5.3 (because get_event_loop was changed to return the event loop of the current task when called from inside a task), and many functions (such as wait_for deprecated the parameter in 3.8 with removal for 3.10. Other functions (like open_connection) didn't deprecate it, but made it optional and documented that it is not needed when called from a coroutine. I wouldn't be surprised if they are deprecated in future.

Only compared to 2.0.0a1:

1. `__del__` will now always use asyncio.get_event_loop(). But `__del__` is broken anyway, for reasons I've explained in #930.
    
2. PubSub.run_in_thread is a synchronous function but needs to know the event loop to run. I've updated it to take an explicit loop argument. This makes it possible to construct the Redis object before the event loop. I'm going to file a separate issue to remove this function anyway, since aioredis should provide concurrency via asyncio tasks rather than threads, and the one test function has been disabled because it is broken.

## Related issue number

#1002

## Checklist

### DIFF
--- a/CHANGES/1002.removal
+++ b/CHANGES/1002.removal
@@ -1,0 +1,2 @@
+Remove explicit loop arguments from connection classes.
+Add loop argument to PubSub.run_in_thread.
--- a/aioredis/client.py
+++ b/aioredis/client.py
@@ -4133,7 +4133,10 @@ class PubSub:
         return message
 
     def run_in_thread(
-        self, daemon: bool = False, exception_handler: Callable = None
+        self,
+        daemon: bool = False,
+        exception_handler: Callable = None,
+        loop: asyncio.AbstractEventLoop = None,
     ) -> "PubSubWorkerThread":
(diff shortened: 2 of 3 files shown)


bmerry:
I'm not sure what to do with regards to testing around PubSub.run_in_thread. redis-py has very minimal testing of this function (just one test for a particular feature), and @seandstewart disabled it in #891 with a comment that it was broken and made the tests hang.

seandstewart (approved):

seandstewart:
Thanks for the work here - it was on my docket to get rid of these so I'm glad it's getting done!

PR #1027, opened 2021-06-15
Set single_connection_client for Pipeline

rad-pat:
## What do these changes do?

Fix `AttributeError` when using a pipeline

## Are there changes in behavior for the user?

No

## Related issue number

https://github.com/aio-libs/aioredis-py/issues/1025

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is `<Name> <Surname>`.
  * Please keep alphabetical order, the file is sorted by names.
- [x] Add a new news fragment into the [`CHANGES/`](../tree/master/CHANGES) folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example:
   `Fix issue with non-ascii contents in doctest text files.`

PR #1029, opened 2021-06-15
add __await__ for Pipeline class

stanislau-arkhipenka:
### TLDR:
without the fix this example:
```
my_key: str = "test_1"
async with redis_client.pipeline() as pipe:
    await pipe.watch(my_key)
    pipe.multi()
    await pipe.set(my_key, 0)
    await pipe.execute()
```
fails with something like this:
```
....
  File "aioredis/client.py", line 879, in initialize
AttributeError: 'Pipeline' object has no attribute 'single_connection_client'
```
### Little bit more details:
The pipeline (pipe) is redis-py's implementation of redis transaction (https://redis.io/topics/transactions).
In terms of aioredis implementation:
`Pipeline `is a child class of `Redis `class
When you try run `pipe.set()` you actually run method from parent (Redis)
inside of this parent's method (pipe.set()) special function `execute_command()` is executed 
This `execute_command()` is overridden in child `Pipeline `and looks something like this:
```
def execute_command(bla):
        if bla:
            return self.immediate_execute_command()
        return self.pipeline_execute_command()
```
The problem, is that `immediate_execute_command `is async function (async def ...) and `pipeline_execute_command `is NOT (def ...).
Developers didn't wanted to have inconsistancy in returns (sometimes awaitable, sometimes just value), so decided to make `pipeline_execute_command `to return self, so when you execute await `pipe.set() `you actually execute _await_. But await is not defined in child class (Pipeline), therefor parent's (Redis's) method executed, which in turn `executes self.initialize()` which reference `self.single_connection_client` which is not defined in child subclass (Pipeline).

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -43,6 +43,7 @@ Michael Käufl
 Nickolai Novik
 Oleg Butuzov
 Oleksandr Tykhonruk
+Pat Buxton
 Pau Freixes
 Paul Becotte
 Paul Colomiets
--- a/README.md
+++ b/README.md
@@ -44,14 +44,14 @@ Benchmarks can be found here:
 ## Contribute
 
 -   Issue Tracker: <https://github.com/aio-libs/aioredis/issues>
--   Google Group: <https://groups.google.com/forum/>\#!forum/aio-libs
+-   Google Group: <https://groups.google.com/g/aio-libs>
 -   Gitter: <https://gitter.im/aio-libs/Lobby>
 -   Source Code: <https://github.com/aio-libs/aioredis>
--   Contributor's guide: devel
+-   Contributor's guide: [devel](docs/devel.md)
 
 Feel free to file an issue or make pull request if you find any bugs or
 have some suggestions for library improvement.
 
 ## License
 
-The aioredis is offered under a [MIT License](license.md).
+The aioredis is offered under a [MIT License](LICENSE).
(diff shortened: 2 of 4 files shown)


Andrew-Chen-Wang:
@stanislau-arkhipenka is this a duplicate of #1027?

stanislau-arkhipenka:
@Andrew-Chen-Wang yeap, seems this solves pretty much same issue. 
In #1027 author sets single_connection_client to False, so no attribute error will occure, when initialze is called. I don't particularry like it as Pipeline should not have a context on connection, as it operates only with connection pool.

Nevertheless both solutions will work :)

rad-pat:
@stanislau-arkhipenka, I think your solution is better myself. Perhaps you can use the test from my https://github.com/aio-libs/aioredis-py/pull/1027 PR to cover the code?

Andrew-Chen-Wang:
Awesome. Thanks for checking out each other's PR. @stanislau-arkhipenka adding the other PR's test would be great!

stanislau-arkhipenka:
Also added other PR to CONTRIBUTORS as I basically ctrl+c ctrl+v the test

seandstewart (changes requested):
Lets make `async_self` -> `_async_self` since it's an implementation detail, not an official API.

seandstewart (approved):
LGTM, thanks for making that change. Go ahead and [add a news artifact](https://aioredis.readthedocs.io/en/latest/devel/#getting-started) to the CHANGES directory and then I think this is good to go.

PR #1076, opened 2021-07-29
Use ContextVar to track Task-local Lock token

### LINKED ISSUE #1040: [2.0] Should Lock default to using contextvars?
I've never used the Lock class in redis-py. But when glancing at the code, I noticed that the default is to store the lock state in thread-local storage (so that two threads can each independently acquire the lock without interfering with each other).

For aioredis, the standard concurrency mechanism is tasks, not threads, so it probably makes sense to default to using `contextvars` (so that two tasks can independently acquire the lock).

abrookins:
## What do these changes do?

The Lock implementation in redis-py uses thread-local storage
so that multiple threads using the same Lock instance can
acquire the Lock from each other.

Thread-local storage ensures that each thread sees a different
token value.

Thread-local storage does not apply in the Task-based concurrency
that asyncio programs use. To achieve a similar effect, we need
to embed a ContextVar instance within each Lock and store the Lock
instance's token within the ContextVar instance. This allows every
Task that uses the same Lock instance to see a different token.

Thus, if both Task A and Task B refer to Lock 1, Task A can "acquire"
Lock 1 and block Task B from acquiring the same Lock until Task A
"releases" the Lock.

NOTE: The Python documentation suggests only storing ContextVar
instances in the top-level module scope due to issues around
garbage collection. That won't work in the current design of
Lock. For lack of a better alternative, and to preserve the
original design of Lock taken from redis-py, we have created
instances of ContextVar within instances of Lock.

Different from redis-py:

1. Users who wish to refer to a Lock's token must do so using `lock.local_token.get()`.
2. We have to drop support for Python 3.6 to use ContextVar. Python 3.6 is EOL 23 Dec 2021. 🤷 

## Related issue number

Fixes #1040.

## Checklist

- [x] Documentation reflects the changes [NOTE: The internal state of the Lock, being the token, is not something current documentation discusses]

### DIFF
--- a/aioredis/lock.py
+++ b/aioredis/lock.py
@@ -1,13 +1,30 @@
 import asyncio
-import threading
+import contextvars
 import uuid
-from types import SimpleNamespace
 from typing import TYPE_CHECKING, Awaitable, NoReturn, Union
 
 from aioredis.exceptions import LockError, LockNotOwnedError
 
 if TYPE_CHECKING:
     from aioredis import Redis
+
+
+class SimpleToken:
+    """
+    Simple storage for a Lock that isn't using task-local storage.
+
+    This class allows us to share the same interface as the ContextVar
+    that a Lock uses for task-local storage.
+    """
+
+    def __init__(self):
+        self.token = ""
+
+    def get(self):
+        return self.token
+
+    def set(self, token: str):
+        self.token = token
 
 
 class Lock:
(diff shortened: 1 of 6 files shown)


aioredis/lock.py
```diff
@@ -1,7 +1,6 @@
 import asyncio
-import threading
+import contextvars
```
Andrew-Chen-Wang, line 2: contextvars isn't supported on pypy, so if we adopt this implementation, then we should simply not support locks for pypy implementations until we can find a solution.
```suggestion
try:
    import contextvars
except ImportError:
    pass
```

aioredis/lock.py
```diff
@@ -289,17 +304,18 @@ def reacquire(self) -> Awaitable[bool]:
...
     async def do_reacquire(self) -> bool:
         timeout = int(self.timeout * 1000)
+        token = self.local_token.get()
         if not bool(
             await self.lua_reacquire(
-                keys=[self.name], args=[self.local.token, timeout], client=self.redis
+                keys=[self.name], args=[token, timeout], client=self.redis
             )
         ):
             raise LockNotOwnedError("Cannot reacquire a lock that's" " no longer owned")
```
Andrew-Chen-Wang, line 321:
```suggestion
            raise LockNotOwnedError("Cannot reacquire a lock that's no longer owned")
```

Andrew-Chen-Wang (changes requested):

Andrew-Chen-Wang:
In the aredis lib, they used a install_requires if condition in the string to install contextvars package (a Python 3.5 port). Not sure if that'll be able to apply in pypy. Would be helpful to note in the README.md at the least

abrookins:
This won't work for the reasons detailed here: https://github.com/aio-libs/aioredis-py/issues/1040#issuecomment-889537152

PR #1099, opened 2021-08-07
Add mypy to CI

### LINKED ISSUE #1098: Enable Mypy
It seems we are exposing types in the library, but many parts are still missing annotations. We should look at filling in the remaining annotations and enable mypy checking in the CI.

If any one wants to volunteer for this before I take a look, feel free to add me to any PRs.

Andrew-Chen-Wang:
## What do these changes do?

Added mypy to linter of CI

One thing to note is if we want to allow failure. Not everyone wants mypy, and PRs could be hindered by them. Running them is great anyway, but maybe not necessary to block anything crucial.

No

## Related issue number

Fixes #1098 and other comments made in the past that I'm not gonna go look for.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,10 @@ jobs:
     - name: Run twine checker
       run: |
         twine check dist/*
+    - name: Run mypy
+      run: |
+        pip install tests/requirements-mypy.txt
+        mypy aioredis --ignore-missing-imports
 
   test-unix:
     name: Test Unix
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
     - name: Run mypy
       run: |
         pip install -r tests/requirements-mypy.txt
-        mypy aioredis
+        mypy aioredis docs/examples
 
   test-unix:
     name: Test Unix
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -49,6 +49,10 @@ jobs:
     - name: Run twine checker
       run: |
         twine check dist/*
+    - name: Run mypy
+      run: |
+        pip install -r tests/requirements-mypy.txt
+        mypy aioredis --ignore-missing-imports
```
Dreamsorcerer, line 55: It's better to just run `mypy` with no arguments, and keep all configuration in `.mypy.ini`.

Good example of a config from one of our other repos:
https://github.com/aio-libs/aiohttp-jinja2/blob/master/.mypy.ini

Andrew-Chen-Wang: Ah just copied our makefile; good catch since we have all dependencies installed. Not sure of the reasoning behind the flag to begin with though.

Dreamsorcerer: But, even the `aioredis` can be removed and just included in the config file as `files = aioredis`.

We'll also want mypy to run on tests/examples at some point as well. We've had bugs with type annotations which are only caught if you type check actual usage (plus it keeps the examples up-to-date with the library code).

Andrew-Chen-Wang: We don't have a config file, so updated in CI only. Maybe in the future.

Andrew-Chen-Wang:
looks like dream sorcerer got it working (prob because member of org). Closing since CI works in #1101.

PR #1101, opened 2021-08-08
Enable mypy in CI

Dreamsorcerer:
Initial mypy config and type fixes to get it enabled in CI.

Starting on this while I'm waiting on some other things.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,0 +1,30 @@
+[mypy]
+#, docs/examples, tests
+files = aioredis
+check_untyped_defs = True
+follow_imports_for_stubs = True
+#disallow_any_decorated = True
+#disallow_any_generics = True
+#disallow_incomplete_defs = True
+disallow_subclassing_any = True
+#disallow_untyped_calls = True
+disallow_untyped_decorators = True
+#disallow_untyped_defs = True
+implicit_reexport = False
+no_implicit_optional = True
+show_error_codes = True
+strict_equality = True
+warn_incomplete_stub = True
+warn_redundant_casts = True
+warn_unreachable = True
+warn_unused_ignores = True
+disallow_any_unimported = True
+#warn_return_any = True
+
+[mypy-aioredis.client]
+# TODO: Fix
+ignore_errors = True
+
+[mypy-aioredis.lock]
+# TODO: Remove once locks has been rewritten
+ignore_errors = True
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -21,10 +21,6 @@ warn_unused_ignores = True
 disallow_any_unimported = True
 #warn_return_any = True
 
-[mypy-aioredis.client]
-# TODO: Fix
-ignore_errors = True
-
 [mypy-aioredis.lock]
 # TODO: Remove once locks has been rewritten
 ignore_errors = True
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -23,6 +23,7 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    cast,
 )
 from urllib.parse import ParseResult, parse_qs, unquote, urlparse
 
@@ -214,7 +215,7 @@ class BaseParser:
     async def can_read(self, timeout: float) -> bool:
         raise NotImplementedError()
 
-    async def read_response(self) -> Union[EncodableT, ResponseError, None]:
+    async def read_response(self) -> Union[EncodableT, ResponseError, None, List[EncodableT]]:
         raise NotImplementedError()
 
 
@@ -516,11 +517,12 @@ class HiredisParser(BaseParser):
                 return False
             raise ConnectionError(f"Error while reading from socket: {ex.args}")
 
-    async def read_response(self) -> EncodableT:
+    async def read_response(self) -> Union[EncodableT, List[EncodableT]]:
         if not self._stream or not self._reader:
             self.on_disconnect()
             raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR) from None
 
+        response: Union[EncodableT, ConnectionError, List[Union[EncodableT, ConnectionError]]]
         # _next_response might be cached from a can_read() call
         if self._next_response is not False:
             response = self._next_response
@@ -538,12 +540,13 @@ class HiredisParser(BaseParser):
         if isinstance(response, ConnectionError):
             raise response
         elif (
-            isinstance(response, list)  # type: ignore[unreachable]
+            isinstance(response, list)
             and response
             and isinstance(response[0], ConnectionError)
         ):
             raise response[0]
-        return response
+        # cast as there won't be a ConnectionError here.
+        return cast(Union[EncodableT, List[EncodableT]], response)
 
 
 DefaultParser: Type[Union[PythonParser, HiredisParser]]
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -754,8 +754,9 @@ class Connection:
 
         # if username and/or password are set, authenticate
         if self.username or self.password:
+            auth_args: Union[Tuple[str], Tuple[str, str]]
             if self.username:
-                auth_args: Tuple[str, ...] = (self.username, self.password or "")
+                auth_args = (self.username, self.password or "")
             else:
                 # Mypy bug: https://github.com/python/mypy/issues/10944
                 auth_args = (self.password or "",)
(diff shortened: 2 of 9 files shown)


seandstewart:
@Dreamsorcerer - 

Thanks for working in this for us! I've added @bmerry as a reviewer, as he's done a lot of work to get our static typing up to snuff.

.mypy.ini
```diff
@@ -0,0 +1,30 @@
+[mypy]
```
Dreamsorcerer, line 1: Commented out things are to be enabled in future PRs. I'll fix up client.py in the next PR.

Dreamsorcerer:
Anybody understand the test failure?

The only change I can see that might result in an error not getting thrown now is:
https://github.com/aio-libs/aioredis-py/pull/1101/files#diff-eebfcf75e78ea2114a6e6019304b6e5709d414bb1e98375a0beafcd593430231L506

According to the hiredis annotations, the code would never get run though...

aioredis/connection.py
```diff
@@ -503,12 +537,6 @@ async def read_response(self) -> EncodableT:
         # happened
         if isinstance(response, ConnectionError):
             raise response
-        elif (
```
seandstewart, line 506: @Dreamsorcerer I'm guessing this is the problem. Add this elif branch back in and the error should now be raised correctly.

Pipelines return a series of responses, not a single response, so we have to inspect the series for an error.

Dreamsorcerer: Ah, I didn't look closely enough at the annotations. It's `self._next_response` which defines the type as `bool`, and the hiredis annotation is `Any`, so fails to highlight the correct issue to us.

Dreamsorcerer: OK, so it checks if it is a list starting with a `ConnectionError`, but what if it's a list with some other value? The return type is `EncodableT` which does not include a list of anything. Should `EncodableT` include a list, or should the return type be something like `Union[EncodableT, List[EncodableT]]`?

seandstewart: At first glance, I'd say do the Union, but iirc that didn't play nicely with the assumptions in the standard client that it's never a list. Union doesn't flag to mypy the case in which a List may occur.

Dreamsorcerer:
client.py is almost done, will probably push it tomorrow.

Could one of you give do a quick review of my PR to upgrade aioredis in aiohttp-session?
https://github.com/aio-libs/aiohttp-session/pull/611

bmerry (commented):
I'm afraid I'm not going to have time to review this, sorry.

Dreamsorcerer:
I wouldn't worry too much about reviewing all the type annotations. Any mistakes should become apparent and be fixed as we expand the typing coverage. If someone can review the minor runtime changes, that's probably all that's really needed before merging this.

I'll wait until you're ready to proceed before making any more progress.

aioredis/connection.py
```diff
@@ -78,7 +82,12 @@
 
 SERVER_CLOSED_CONNECTION_ERROR = "Connection closed by server."
 
-SENTINEL = object()
+
+class Sentinel(enum.Enum):
+    sentinel = object()
```
Andrew-Chen-Wang, line 87: Why is this `enum.Enum` needed? And can you prepend an underscore to `Sentinel`?

aioredis/connection.py
```diff
@@ -233,10 +251,12 @@ def length(self):
     async def _read_from_socket(
         self,
         length: Optional[int] = None,
-        timeout: Optional[float] = SENTINEL,  # type: ignore
+        timeout: Union[float, None, Sentinel] = SENTINEL,
```
Andrew-Chen-Wang, line 254:
```suggestion
        timeout: Optional[Union[float, Sentinel]] = SENTINEL,
```

aioredis/connection.py
```diff
@@ -405,6 +437,7 @@ async def read_response(self) -> Union[EncodableT, ResponseError, None]:
                 return None
             response = [(await self.read_response()) for _ in range(length)]
         if isinstance(response, bytes):
+            assert self.encoder is not None
```
Andrew-Chen-Wang, line 440: Why is this needed? If anything, we should probably raise RedisError instead.

aioredis/connection.py
```diff
@@ -715,9 +756,10 @@ async def on_connect(self):
         # if username and/or password are set, authenticate
         if self.username or self.password:
             if self.username:
-                auth_args = (self.username, self.password or "")
+                auth_args: Tuple[str, ...] = (self.username, self.password or "")
```
Andrew-Chen-Wang, line 759: Intentional `...`?

Andrew-Chen-Wang (approved):
Looks fine to me. Most of this code is just typing, but there are some stuff that isn't; for example, the "cast"-ing is something I'm not really happy with. Or an assertion statement that should instead be an if condition.

Besides that, I think this PR doesn't affect the code base too much.

Though, when it comes to maintenance and PR adding, I do think the CI could be non-blocking like codecov. It'll obviously break people's mypy every so often, but I guess it's a balancing act.

aioredis/connection.py
Dreamsorcerer: Not sure what the likelihood of this being `None` is. If you're confident it should never happen here, then an assert statement makes that clear (while telling mypy that the below code will never be run when it is `None`). If there's a chance that it might be `None` here, then we should raise `RedisError` instead.

aioredis/connection.py
Dreamsorcerer: Indicates a tuple of any length. Given the alternative form on line 762, it's either this or `Union[Tuple[str], Tuple[str, str]]`, which is technically more accurate, but not enough benefit to warrant the awkwardness.

Andrew-Chen-Wang: it's a dumb nitpick of mine, but I'd like the more verbose `Union[Tuple[str], Tuple[str, str]]`. Thanks for being patient with me.

aioredis/connection.py
Andrew-Chen-Wang: Hmm IIRC, we've had an issue with the buffer in the Encoder class before, but the Encoder object itself I've never seen become `None` before. So I think it's fine to remove the assertion.

Dreamsorcerer:
>  for example, the "cast"-ing is something I'm not really happy with.

Note that casting is just typing. At runtime it just returns the argument unchanged.
Although I do try to avoid casting unless the only other option is to ignore the errors.

> Though, when it comes to maintenance and PR adding, I do think the CI could be non-blocking like codecov. It'll obviously break people's mypy every so often, but I guess it's a balancing act.

I think the problem is that the easiest time to fix a typing error is when you make the change. i.e. The author of a PR should fix the errors in that PR.
For example, me trying to fix the typing errors now is a lot trickier as I have to keep making assumptions and guesses at the intended behaviour. I'm sure there are some typing mistakes in there which will be revealed through later PRs when the typing is expanded.

aioredis/connection.py
Dreamsorcerer: Relooking at it, I could make the check at the top of the function be `if not self._buffer or not self.encoder:`. That could also work, as it ensures encoder is not `None` and resolves the mypy error. The only code I see that sets it to `None` is the disconnect, so it's essentially the same check (although it's a public attribute, which suggests a user could assign `None` directly).

Andrew-Chen-Wang:
> Note that casting is just typing. At runtime it just returns the argument unchanged.
Although I do try to avoid casting unless the only other option is to ignore the errors.

I was just thinking it as an extra, "unnecessary" operation in terms of actual execution of commands.

> I think the problem is that the easiest time to fix a typing error is when you make the change. i.e. The author of a PR has the best knowledge of that code during the period of working on that PR.

That's true. @abrookins Can you provide a review of in general 1. should we still include the mypy CI (I approve) and 2. any other concerns? I wouldn't bother reviewing the entire thing since we can just change annotations over time if there is something incorrect.

aioredis/connection.py
Dreamsorcerer: Sentinels are a little messy to type. It seems that an enum is the best approach to use:
https://stackoverflow.com/a/60605919

Andrew-Chen-Wang:
Thanks a lot @Dreamsorcerer !

Dreamsorcerer:
Oh man, I'm trying to add the examples to the type checking and aioredis doesn't lend itself to static type checking very well.

The examples use `Pipeline` which inherits from `Redis`. However, `Redis.get()` etc. return an `Awaitable`, but when using the same method on `Pipeline`, it may return a `Pipeline`. This violates the Liskoff Substitution Principle and makes it very hard, maybe impossible, to type it correctly. The return type on `Pipeline.execute_command()` is also `Union[Pipeline, Awaitable[Pipeline]]`, which is basically useless to the user as they will always get an error when using it.

PR #1129, opened 2021-09-16
fix socket.error raises

### LINKED ISSUE #1121: ConnectionResetError: [WinError 10054] 远程主机强迫关闭了一个现有的连接。
idk what happend, but it dosen't work suddenly.
i guarantee redis server is ok and i try other redis lib it works too,

i am use version:
aioredis==2.0.0
  - async-timeout [required: Any, installed: 3.0.1]
  - typing-extensions [required: Any, installed: 3.10.0.2]
 
on Microsoft Windows [Version 10.0.19043.1165]
python 3.9.6

x0day:
Fixes #1121

## What do these changes do?

fix raise `aioredis.exceptions.ConnectionError` instead of raw `socket.error` or its subclasses

no

## Related issue number

#1121


eneloop2:
This kind of bug appears in more than one place so it's not actually going to address all instances of `ConnectionRestError` being raised.

x0day:
@eneloop2 @Andrew-Chen-Wang 
https://github.com/aio-libs/aioredis-py/issues/1121#issuecomment-920932796
look at this, please consider this fix, thanks.

```
from aioredis.connection import NONBLOCKING_EXCEPTIONS

try:
    raise ConnectionResetError
except NONBLOCKING_EXCEPTIONS:
    print('aioredis catched!')
except Exception as exc:
    print("aioredis not catched!")

from redis.connection import NONBLOCKING_EXCEPTIONS

try:
    raise ConnectionResetError
except NONBLOCKING_EXCEPTIONS:
    print('redis-py catched')
```

Andrew-Chen-Wang:
@x0day do you have hiredis installed by chance?

x0day:
> @x0day do you have hiredis installed by chance?

YES, it's not parse error, just aio-redis not consider socket error as NONBLOCKING_EXCEPTIONS.

x0day:
@seandstewart  can we get this in the PR #1156 ?

Andrew-Chen-Wang (approved):
Thanks for the message and PR!

Andrew-Chen-Wang:
Please add a CHANGE file though

x0day:
@Andrew-Chen-Wang thanks for the review!

Andrew-Chen-Wang:
One thing to note is that from https://github.com/redis/redis-py/commit/8c5a41baf0bd2a1388d601e5b49d06b91997ccb8, this piece of code was dropped due to Python 2.7 dropping. How does this resolve a Windows specific error?

x0day:
@Andrew-Chen-Wang 

FYI: at redis/redis-py@8c5a41b, it's solve the problem by raise `ConnectionError` instead of raw `socket.error`, aioredis-py's is same here.

image

but aioredis-py didn't catch the `OSError` at the `Connection.read_response`. it's the difference. 

https://github.com/redis/redis-py/blob/791f482dcb320f48cf950c4b2c6047d1981a8f67/redis/connection.py#L753-L756
https://github.com/aio-libs/aioredis-py/blob/2ba15fb6947fa2347d401ba436e362ad62ed38ff/aioredis/connection.py#L893-L907

so if `socket.error` in `NONBLOCKING_EXCEPTIONS`, `ConnectionResetError` will be catched at `_read_from_socket `, either handle the `OSError` in `Connection.read_response` also solve the bug.

should I need to update the PR, by catching the `OSError` in `Connection.read_response` like the redis-py implement？

Andrew-Chen-Wang:
Yes, by catching the socket.error you mean? I'm surprised BaseException doesn't catch it, but at least by catching socket.error directly we can resolve this issue + be in paralle l with redis-py.

x0day:
@Andrew-Chen-Wang can you review the PR again? thanks.

CHANGES/1129.bugfix
```diff
@@ -0,0 +1 @@
+fixed raw socket.error(or one of its subclasses) raises instead of a redis.exceptions.ConnectionError
```
Andrew-Chen-Wang, line 1:
```suggestion
fixed raw socket.error (or one of its subclasses) raises instead of a redis.exceptions.ConnectionError
```

aioredis/connection.py
```diff
@@ -858,6 +858,11 @@ async def read_response(self):
         except asyncio.TimeoutError:
             await self.disconnect()
             raise TimeoutError(f"Timeout reading from {self.host}:{self.port}")
+        except OSError as e:
```
Andrew-Chen-Wang, line 861: Seems too overarching. Is there a reason not to catch socket.error and socket.SOMETHING (assuming IIRC redis-py caught two errors from socket) specifically?

Andrew-Chen-Wang (changes requested):

aioredis/connection.py
x0day: yes, you are right!

x0day: https://docs.python.org/3.6/library/socket.html#socket.error

socket.error has been deprecated, socket.error is an alias for `OSError`

Andrew-Chen-Wang:
failing ci

x0day:
@Andrew-Chen-Wang sorry for late response, failing CI because this:

Andrew-Chen-Wang (approved):
not sure to be honest; maybe to decrease imports? If anything, LGTM

PR #1207, opened 2021-11-17
Fix health check after unsubscribe

### LINKED ISSUE #1206: Health check fails when a pubsub has no subscriptions
### Describe the bug

When a PubSub needs to issue a PING due to the health check feature, it does not consider that there might be no subscriptions at the moment. Redis responds differently to PING depending on whether there are active subscriptions or not: if there are no subscriptions it just returns the argument as a bulk response, instead of a multi-bulk with "pong" and the response. This breaks the code that detects the health check response, and instead the individual bytes of the `aioredis-py-health-check` string get inserted into the returned message.

### To Reproduce

1. Install aioredis 2.0.0
2. Run this code:
```python
#!/usr/bin/env python3

import asyncio

import aioredis


async def poll(ps):
    while True:
        message = await ps.get_message(timeout=1)
        if message is not None:
            print(message)


async def main():
    r = aioredis.Redis.from_url("redis://localhost", health_check_interval=2)
    ps = r.pubsub()
    await ps.subscribe("foo")
    poller = asyncio.create_task(poll(ps))
    await asyncio.sleep(5)
    await ps.unsubscribe("foo")
    await asyncio.sleep(5)
    await ps.subscribe("baz")
    poller.cancel()
    try:
        await poller
    except asyncio.CancelledError:
        pass

asyncio.run(main())
```

### Expected behavior

Expected all messages printed to have proper types.

### Logs/tracebacks

```python-traceback
{'type': 'subscribe', 'pattern': None, 'channel': b'foo', 'data': 1}
{'type': 'unsubscribe', 'pattern': None, 'channel': b'foo', 'data': 0}
{'type': 97, 'pattern': None, 'channel': 105, 'data': 111}
{'type': 97, 'pattern': None, 'channel': 105, 'data': 111}
```

Note that 97, 105, 111 are the result of indexing b"aioredis-py-health-check" with indices 0, 1, 2.
```


### Python Version

```console
$ python --version
Python 3.8.10
```

### aioredis Version

```console
$ python -m pip show aioredis
Name: aioredis
Version: 2.0.0
```

### Additional context

redis-py seems to have a [similar bug](https://github.com/redis/redis-py/issues/1720) with the interaction between health checks and pub-sub, but the failure mode is not the same (in redis-py it seems to be some sort of race condition, whereas in aioredis it appears reliably reproducible), so this might need an aioredis-specific fix.

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Andrew-Chen-Wang:
## What do these changes do?

Returns None for health check response.

Maybe. If someone used to check the health check response after unsubscribing (for some odd reason), then yes. But for most people, no. And I can't imagine why since we usually return None for health checks.

## Related issue number

Fixes #1206

## Checklist

- [X] I think the code is well written
- [X] Documentation reflects the changes
- [X] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

### DIFF
--- a/aioredis/client.py
+++ b/aioredis/client.py
@@ -3937,16 +3937,14 @@ class PubSub:
         # we need to know the encoding options for this connection in order
         # to lookup channel and pattern names for callback handlers.
         self.encoder = self.connection_pool.get_encoder()
+        self.health_check_message_b = self.encoder.encode(self.HEALTH_CHECK_MESSAGE)
         if self.encoder.decode_responses:
             self.health_check_response: Iterable[Union[str, bytes]] = [
                 "pong",
                 self.HEALTH_CHECK_MESSAGE,
             ]
         else:
-            self.health_check_response = [
-                b"pong",
-                self.encoder.encode(self.HEALTH_CHECK_MESSAGE),
-            ]
+            self.health_check_response = [b"pong", self.health_check_message_b]
         self.channels: Dict[ChannelT, PubSubHandler] = {}
         self.pending_unsubscribe_channels: Set[ChannelT] = set()
         self.patterns: Dict[ChannelT, PubSubHandler] = {}
@@ -4049,7 +4047,10 @@ class PubSub:
             return None
         response = await self._execute(conn, conn.read_response)
 
-        if conn.health_check_interval and response == self.health_check_response:
+        if (
+            conn.health_check_interval
+            and response in (self.health_check_response, self.health_check_message_b)
+        ):
             # ignore the health check message as user might not expect it
             return None
         return response


seandstewart (approved):

bmerry:
The code LGTM for this specific issue, but it is probably worth adding a comment explaining why you're checking for two possible responses.

aioredis/client.py
```diff
@@ -4049,7 +4047,10 @@ async def parse_response(self, block: bool = True, timeout: float = 0):
             return None
         response = await self._execute(conn, conn.read_response)
 
-        if conn.health_check_interval and response == self.health_check_response:
+        if conn.health_check_interval and response in (
+            self.health_check_response,
+            self.health_check_message_b,
+        ):
```
bmerry, line 4053:
```suggestion
        # The response depends on whether there were any subscriptions
        # active at the time the PING was issued.
        if conn.health_check_interval and response in (
            self.health_check_response,  # If there was at least one subscription
            self.health_check_message_b,  # If there wasn't
        ):
```

bmerry (changes requested):

bmerry (approved):

Andrew-Chen-Wang:
not sure what's going on. Will come back later if no one else figures out what's wrong. Branch should be open for anyone to change...

Issue #1114, opened 2021-08-28
2.0.0: pytest is failing


kloczek:
I'm trying to package your module as an rpm package. So I'm using the typical build, install and test cycle used on building packages from non-root account.
- "setup.py build"
- "setup.py install --root </install/prefix>"
- "pytest with PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>

May I ask for help? One unit is failing.
```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-aioredis-2.0.0-2.fc35.x86_64/usr/lib64/python3.8/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-aioredis-2.0.0-2.fc35.x86_64/usr/lib/python3.8/site-packages
+ /usr/bin/pytest -ra
934942:C 28 Aug 2021 13:45:54.042 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
934942:C 28 Aug 2021 13:45:54.042 # Redis version=6.2.2, bits=64, commit=00000000, modified=0, pid=934942, just started
934942:C 28 Aug 2021 13:45:54.042 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
934942:M 28 Aug 2021 13:45:54.043 * Increased maximum number of open files to 10032 (it was originally set to 1024).
934942:M 28 Aug 2021 13:45:54.043 * monotonic clock: POSIX clock_gettime
                _._
           _.-``__ ''-._
      _.-``    `.  `_.  ''-._           Redis 6.2.2 (00000000/0) 64 bit
  .-`` .-```.  ```\/    _.,_ ''-._
 (    '      ,       .-`  | `,    )     Running in standalone mode
 |`-._`-...-` __...-.``-._|'` _.-'|     Port: 6379
 |    `-._   `._    /     _.-'    |     PID: 934942
  `-._    `-._  `-./  _.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |           https://redis.io
  `-._    `-._`-.__.-'_.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |
  `-._    `-._`-.__.-'_.-'    _.-'
      `-._    `-.__.-'    _.-'
          `-._        _.-'
              `-.__.-'

[...]

Issue #1212, opened 2021-11-20
Exception RedisError("Buffer is closed") when trying to use library with PythonParser


m-novikov:
### Describe the bug

When using PythonParser is used attempt to call command raises an exception

### To Reproduce

```python
import aioredis
import aioredis.connection

async def run():
    redis = aioredis.from_url("redis://localhost:6379", parser_class=aioredis.connection.PythonParser)
    await redis.ping()

if __name__ == "__main__":
    asyncio.run(run())
```

### Expected behavior

No exception is raised

### Logs/tracebacks

```python-traceback
Traceback (most recent call last):
  File "/home/novikov/src/aiohttp-bench/aioredis-py/bench.py", line 35, in <module>
    asyncio.run(run2())
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/novikov/src/aiohttp-bench/aioredis-py/bench.py", line 31, in run2
    await pool.ping()
  File "/home/novikov/src/aiohttp-bench/aioredis-py/aioredis/client.py", line 1079, in execute_command
    conn = self.connection or await pool.get_connection(command_name, **options)
  File "/home/novikov/src/aiohttp-bench/aioredis-py/aioredis/connection.py", line 1410, in get_connection
    await connection.connect()
  File "/home/novikov/src/aiohttp-bench/aioredis-py/aioredis/connection.py", line 703, in connect
    await self.on_connect()
  File "/home/novikov/src/aiohttp-bench/aioredis-py/aioredis/connection.py", line 755, in on_connect
    self._parser.on_connect(self)
  File "/home/novikov/src/aiohttp-bench/aioredis-py/aioredis/connection.py", line 377, in on_connect
    raise RedisError("Buffer is closed.")
aioredis.exceptions.RedisError: Buffer is closed.
```

### Python Version
3.9

### aioredis Version

```console
master at 33b2dbd0a40ac148e6a36ba2fc7ab5d438a9a71d
```

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

PR #1213, opened 2021-11-20
Fix buffer is closed error when using PythonParser class

m-novikov:
## What do these changes do?
Fix issue when the exception is raised on connection initialization in PythonParser
Add parametrization for both parsers to all tests using redis client to catch these errors in the future

No

## Related issue number
#1212
#1114

## Checklist

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -373,7 +373,7 @@ class PythonParser(BaseParser):
     def on_connect(self, connection: "Connection"):
         """Called when the stream connects"""
         self._stream = connection._reader
-        if self._buffer is None or self._stream is None:
+        if self._stream is None:
             raise RedisError("Buffer is closed.")
 
         self._buffer = SocketBuffer(
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -4,17 +4,15 @@ from unittest import mock
 
 import pytest
 
-from aioredis.connection import UnixDomainSocketConnection
+from aioredis.connection import PythonParser, UnixDomainSocketConnection
 from aioredis.exceptions import InvalidResponse
-from aioredis.utils import HIREDIS_AVAILABLE
-
-if TYPE_CHECKING:
-    from aioredis.connection import PythonParser
 
 
-@pytest.mark.skipif(HIREDIS_AVAILABLE, reason="PythonParser only")
 @pytest.mark.asyncio
-async def test_invalid_response(r):
+@pytest.mark.parametrize("create_redis", [(True, PythonParser)], indirect=True)
+async def test_invalid_response(create_redis):
+    r = await create_redis()
+
     raw = b"x"
     parser: "PythonParser" = r.connection._parser
     with mock.patch.object(parser._buffer, "readline", return_value=raw):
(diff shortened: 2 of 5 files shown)


aioredis/connection.py
```diff
@@ -373,7 +373,7 @@ def __init__(self, socket_read_size: int):
     def on_connect(self, connection: "Connection"):
         """Called when the stream connects"""
         self._stream = connection._reader
-        if self._buffer is None or self._stream is None:
+        if self._stream is None:
```
m-novikov, line 376: I am not too sure what was the original intent behind this line, buffer only initialized after so it only can be `None` in this line.

Andrew-Chen-Wang: i believe we've had unexpected behavior here before, so keeping that there would be nice

m-novikov: `self._stream is None` check is kept and test seems to be passing with it, The problem was `self._buffer is None` which always resulted in an exception. As it's not possible for it to be something other than `None` at this point.

tests/test_connection.py
```diff
@@ -4,17 +4,15 @@
...
 
-if TYPE_CHECKING:
-    from aioredis.connection import PythonParser
 
-
-@pytest.mark.skipif(HIREDIS_AVAILABLE, reason="PythonParser only")
 @pytest.mark.asyncio
-async def test_invalid_response(r):
+@pytest.mark.parametrize("create_redis", [(True, PythonParser)], indirect=True)
+async def test_invalid_response(create_redis):
```
m-novikov, line 13: This test seems like it wasn't run for some time, as in the case of connection pool it will fail with `AttributeError`, because only a single connection client has a connection set

m-novikov: And it also didn't work on py =< 3.7

Andrew-Chen-Wang: > And it also didn't work on py =< 3.7

We should probably drop Python 3.6 now anyways since it's EOL soon.

m-novikov: But for 3.7 it's still a problem and pypy seems to be unhappy about this test also.
AsyncMock solves this problem.

Andrew-Chen-Wang (approved):
This LGTM. Change was in the mypy integration in #1101 Thanks!

Andrew-Chen-Wang:
@m-novikov are the changes to the test files necessary though or is it possible to simply delete the if condition for `self._buffer`?

m-novikov:
@Andrew-Chen-Wang  changes in tests files are necessary to ensure that test pass also with `PythonParser` it seems like CI pipeline didn't actually run against `PythonParser` as `HiredisParser` is available. At least from my understanding, this also closes #1114

m-novikov:
I am not actually sure if it worth to test all combinations, or only 3 of them should suffice. e.g. (hiredis pool and single) and python parser single

Andrew-Chen-Wang:
I agree that #1114 can be closed. I think having everything being tested is nice, though on CLI it might be annoying if you don't want to wait a full minute for the test suite to finish.

Andrew-Chen-Wang:
Let's merge this full test suite solution first; a separate PR for the marker solution would be great. I'll direct the other maintainers to decide on that.

Andrew-Chen-Wang:
Thanks again!

Issue #1208, opened 2021-11-19
aioredis 1.3.1 vs 2.0.0 performance


artesby:
### Describe the bug

aioredis==2.0.0 performs x4-10 times slower than aioredis==1.3.1

to see this we did a simple  test:

<details> 
<summary>test_aioredis1.py</summary>

```python
import asyncio
import os
import time
import aioredis

url = os.environ.get('REDIS_URL', 'redis://localhost:6379/13')
max_conn = os.environ.get('MAX_CONNECTIONS', 64)

async def task(i, redis):
    key = f'key:{i}'
    v = await redis.get(key)
    new_v = 1 if v is None else int(v) + 1
    await redis.set(key, new_v, expire=600)

async def run(n=1500):
    redis = await aioredis.create_redis_pool(
        url, encoding='utf-8', maxsize=max_conn
    )

    tasks = [asyncio.create_task(task(i, redis)) for i in range(n)]
    start = time.time()
    await asyncio.gather(*tasks)
    t = time.time() - start
    print(f'aioredis1: {n} tasks with blocking pool with {max_conn} connections: {t}s')


if __name__ == "__main__":
    asyncio.run(run(n=1500))

```
</details> 

<details> 
<summary>test_aioredis2.py</summary>

```python
import asyncio
import os
import time
from aioredis import BlockingConnectionPool, Redis

url = os.environ.get('REDIS_URL', 'redis://localhost:6379/14')
max_conn = os.environ.get('MAX_CONNECTIONS', 64)

async def task(i, redis):
    key = f'key:{i}'
    v = await redis.get(key)
    new_v = 1 if v is None else int(v.decode()) + 1
    await redis.set(key, new_v, ex=600)

async def run(n=1500):
    pool = BlockingConnectionPool.from_url(
        url=url, max_connections=max_conn
    )
    redis = Redis(connection_pool=pool)

    tasks = [asyncio.create_task(task(i, redis)) for i in range(n)]
    start = time.time()
    await asyncio.gather(*tasks)
    t = time.time() - start
    print(f'aioredis2: {n} tasks with blocking pool with {max_conn} connections: {t}s')


if __name__ == "__main__":
    asyncio.run(run(n=1500))
```
</details> 

### To Reproduce

1. pip3 install aioredis==2.0.0 && python3 test_aioredis2.py
2. pip3 install aioredis==1.3.1 && python3 test_aioredis1.py0

### Expected behavior

[...]

PR #1216, opened 2021-11-21
[RFC] Use callback based asyncio.Protocol for communication

m-novikov:
## What do these changes do?

Attempt to refactor connection code to avoid unnecessary polling and waiting. 

When testing this branch use a single connection client, connection pool has some locking around it that need to be ironed out.
But performance wise pool doesn't actually help, for [simplistic client](https://github.com/m-novikov/aioredis-benchmarks/blob/main/aioredis_benchmarks/baseline.py)
```python
async with Redis(connection_pool=pool).client() as redis:
    ...
```

Also this PR uses only `HiredisParser` because `feed` parser API is needed to conveniently use with protocol callbacks.
In case this approach is feasible `PythonParser` would also need to be refactored(rewritten?) to support `feed`.

## Related issue number

#1208

## Checklist

### DIFF
--- a/aioredis/connection.py
+++ b/aioredis/connection.py
@@ -8,6 +8,7 @@ import socket
 import ssl
 import threading
 import warnings
+from collections import deque
 from distutils.version import StrictVersion
 from itertools import chain
 from types import MappingProxyType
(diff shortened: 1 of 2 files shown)


Andrew-Chen-Wang:
@m-novikov do you have time to resolve the mypy errors?

m-novikov:
Ah, I though to get some feedback about idea before proceeding :) it requires more work to pass the tests, if you think idea is worthwhile I can

Andrew-Chen-Wang:
I'm not that familiar with Protocol and Transports, though the code looks like event driven languages? I've cc-ed @bmerry to give a quick look at it, but, aside from my personal concern about instability and inconsistencies (I could be wrong since again I'm unfamiliar with this) in response data ordering, I'd say it's worth a shot! 🚀

aioredis/connection.py
```diff
@@ -552,6 +553,105 @@ async def read_response(self) -> Union[EncodableT, List[EncodableT]]:
...
+        self._state = _State.not_connected
+        self._resp_queue = deque()
+
+        self._transport: Optional[asyncio.Transport] = None
+
+        self._parser = hiredis.Reader()
+
+        self._responder_task = None
+        self._exc = None
+        self._conn_waiter = asyncio.get_event_loop().create_future()
```
bmerry, line 576: I'd recommend using an `asyncio.Event` for this. With a raw future, calling `wait_connected` and then cancelling it will cancel the future. Although it might not be needed at all - see my comments where `create_connection` is called.

aioredis/connection.py
```diff
@@ -552,6 +553,105 @@ async def read_response(self) -> Union[EncodableT, List[EncodableT]]:
...
+            # It's possible that connection was dropped and connection_callback was not
+            # called yet, to stop spamming errors, avoid writing to broken pipe
+            # Both _UnixWritePipeTransport and _SelectorSocketTransport that we
+            # expect to see here have this attribute
+            if not self._transport._conn_lost:
+                self._transport.write(cmd)
+            self._resp_queue.append(fut)
+
+        elif self._state == _State.not_connected:
+            fut.set_exception(Exception("Not connected"))
```
bmerry, line 607: Is there a more appropriate subclass e.g. ConnectionError?

aioredis/connection.py
```diff
@@ -552,6 +553,105 @@ async def read_response(self) -> Union[EncodableT, List[EncodableT]]:
...
+
+        elif self._state == _State.error:
+            assert self._exc
+            fut.set_exception(self._exc)
+
+        return fut
+
+    def data_received(self, data):
+        if self._state != _State.connected:
+            return
```
bmerry, line 617: Should this be an exception? Or is this something that happens if the client disconnects and we receive data that was still in the socket?

aioredis/connection.py
```diff
@@ -552,6 +553,105 @@ async def read_response(self) -> Union[EncodableT, List[EncodableT]]:
...
+    def _set_exception(self, exc):
+        self._exc = exc
+        self._state = _State.error
+
+    def connection_lost(self, exc):
+        if exc is not None:
+            self._set_exception(exc)
+
+        elif self._state == _State.connected:
+            exc = Exception("disconnected")
```
bmerry, line 644: Should probably be a more specific subclass of Exception.

aioredis/connection.py
```diff
@@ -552,6 +553,105 @@ async def read_response(self) -> Union[EncodableT, List[EncodableT]]:
...
+
+        self._responder_task = None
+        self._exc = None
+        self._conn_waiter = asyncio.get_event_loop().create_future()
+
+    def connection_made(self, transport: asyncio.Transport) -> None:
+        self._transport = transport
+        sock = transport.get_extra_info("socket")
+        if sock is not None:
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
```
bmerry, line 582: The original code also tries to set KEEPALIVE.

aioredis/connection.py
```diff
@@ -716,28 +818,30 @@ async def connect(self):
...
-                    sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
-                    for k, v in self.socket_keepalive_options.items():
-                        sock.setsockopt(socket.SOL_TCP, k, v)
-
-            except (OSError, TypeError):
-                # `socket_keepalive_options` might contain invalid options
-                # causing an error. Do not leave the connection open.
-                writer.close()
-                raise
+            await self._protocol.wait_connected()
```
bmerry, line 827: Is this necessary? Looking at the documentation for create_connection it looks like it will call `connection_made` on the protocol before returning.

aioredis/connection.py
```diff
@@ -552,6 +553,105 @@ async def read_response(self) -> Union[EncodableT, List[EncodableT]]:
...
+        await self._conn_waiter
+
+    def send_command(self, cmd):
+        fut = asyncio.get_event_loop().create_future()
+
+        if self._state == _State.connected:
+            # It's possible that connection was dropped and connection_callback was not
+            # called yet, to stop spamming errors, avoid writing to broken pipe
+            # Both _UnixWritePipeTransport and _SelectorSocketTransport that we
+            # expect to see here have this attribute
```
bmerry, line 601: What about _ProactorSocketTransport? I don't know Windows so don't really know anything about it.

bmerry (commented):
I like the concept. Have you tried it on pub/sub yet? I'm not sure how well the change to have `send_command` return a response future will work with the pub/sub code, which
(a) doesn't immediately retrieve responses to subscribe commands, instead leaving them for the message handler to pick up
(b) needs to be able to handle unsolicited messages.

Possibly the Protocol could just put responses into an asyncio.Queue, and `read_response` could then pull a response from the queue. If the queue gets too full (which might just be 2 responses), use `pause_reading` to stop more data being added until it shrinks again. But I don't know if that will preserve the ordering one needs to make pipelining work, and it might also reduce performance again.

Do you have any thoughts on implementing the `pause_writing` and `resume_writing` callbacks? I can't think of anything useful they could do, since there is no way to apply back-pressure to a client that is sending lots of commands without awaiting responses.

I'm assuming the commented-out code is things you need to revisit before finalising things.

bmerry:
By the way I am going on an extended holiday soon, so might not be able to give further feedback.

seandstewart:
superseded by #1287

PR #1224, opened 2021-11-30
Fix typing on blpop (etc) timeout argument

icgood:
## What do these changes do?

Improves typing on [blpop](https://redis.io/commands/blpop)'s *timeout* argument, which is a number of seconds (not milliseconds) and allows decimal precision:

> The timeout argument is interpreted as a double value specifying the maximum number of seconds to block. A timeout of zero can be used to block indefinitely.

Other blocking `b*` methods were updated similarly. I also fixed some low-hanging mypy errors under `tests/`.

No change in behavior, only type hint changes.

## Related issue number

N/A

## Checklist

### DIFF
--- a/CHANGES/1224.bugfix
+++ b/CHANGES/1224.bugfix
@@ -1,0 +1,1 @@
+Fix typing on blpop (etc) timeout argument
--- a/aioredis/client.py
+++ b/aioredis/client.py
@@ -65,6 +65,7 @@ GroupT = _StringLikeT  # Consumer group
 ConsumerT = _StringLikeT  # Consumer name
 StreamIdT = Union[int, _StringLikeT]
 ScriptTextT = _StringLikeT
+TimeoutSecT = Union[int, float, _StringLikeT]
 # Mapping is not covariant in the key type, which prevents
 # Mapping[_StringLikeT, X from accepting arguments of type Dict[str, X]. Using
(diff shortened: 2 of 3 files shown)


Andrew-Chen-Wang (changes requested):
Thanks! We'll want to change the name of the TimeoutSecT to something more specific to BL/BR ops because there are other arguments that also include a timeout argument.

icgood:
> We'll want to change the name of the TimeoutSecT to something more specific to BL/BR ops because there are other arguments that also include a timeout argument.

@Andrew-Chen-Wang Yeah, that was the reason I went with `TimeoutSecT`, to differentiate it from other `timeout=` arguments that take milliseconds. I verified that the spots I switched to `TimeoutSecT` were all documented as taking seconds, the same way as `BLPOP`. Would any of the following be preferable to you?

* `TimeoutSecondsT`
* `SecondsT`
* `TimeoutSecondsT` + `TimeoutMillisT` for millisecond values (simple alias for `int`).

Andrew-Chen-Wang:
I see. I fear that some other redis command would also have seconds for timeout arg. I was thinking in terms of blocking list like `BLTimeoutT`, but if the only redis commands that use `seconds`are the blocking list primitives, then I think the current state is fine.

Andrew-Chen-Wang (approved):
Thanks a lot!

PR #1227, opened 2021-12-02
Remove __del__ from Redis (Fixes #1115)

### LINKED ISSUE #1115: Warning caused by `Redis.__del__()`
This method calls `create_task()`, but then never awaits on that task. This raises a warning in Python, which fails CI for aiohttp-session. This happens even if we've closed the client explicitly.

I was able to fix the warning by patching out the `__del__()` method: https://github.com/aio-libs/aiohttp-session/pull/616/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R21

Explicit is better than implicit, so I'd argue that this method doesn't need to exist, when you already have both a `.close()` method and support for `async with`. It's also misleading and a potential performance issue if an async client suddenly starts blocking on IO implicitly. Compare aiohttp's client, which merely raises an error:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L284-L299

Andrew-Chen-Wang:
## What do these changes do?

Removes `__del__` from `Redis`

Yes, you should explicitly disconnect and not use `__del__`

## Related issue number

Fixes #1115

## Checklist


Dreamsorcerer:
I'm fine with this, but better would be to copy the code in aiohttp to raise a warning:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L294-L309

bmerry (commented):
I don't have anything to add to what @Dreamsorcerer has said.

aioredis/client.py
```diff
@@ -1054,14 +1054,23 @@ async def __aenter__(self: _RedisT) -> _RedisT:
...
+            if self.connection is not None:
+                _warnings.warn(
+                    f"Unclosed client session {self!r}",
+                    ResourceWarning,
+                    source=self,
+                )
+                context = {"client": self, "message": self._DEL_MESSAGE}
+                if self._source_traceback is not None:
+                    context["source_traceback"] = self._source_traceback
+                self._loop.call_exception_handler(context)
```
Dreamsorcerer, line 1070: Not sure what `_source_traceback` is, so maybe can just drop those 2 lines.
Then `self._loop` will probably need replacing with `asyncio.get_running_loop()`.

aioredis/client.py
```diff
@@ -1054,14 +1054,23 @@ async def __aenter__(self: _RedisT) -> _RedisT:
...
+        except AttributeError:
+            # loop was not initialized yet,
+            # either self._connector or self._loop doesn't exist
```
Dreamsorcerer, line 1073: Looks like this is also irrelevant, so should be able to drop the `try ... except`.

Andrew-Chen-Wang:
@bmerry @seandstewart sorry for the spam requests for review. After abrookins left, I'm wondering who still currently has power for approving PRs to be merged?

seandstewart (approved):

seandstewart:
@Andrew-Chen-Wang - I believe you should have power to merge this. LGTM 🚢

Andrew-Chen-Wang:
Thanks @seandstewart . To clarify, PRs require authorized reviewers to approve PRs before they get merged. Just wondering besides us two, who else has the privilege to approve PRs since I feel guilty spam requesting reviews 😅

Issue #1235, opened 2021-12-13
subscription callback function cannot be coroutine or awaitable


dnlwgnd:
### Describe the bug

I am using aioredis in a FastAPI app. 
I want to use the pubsub feature and setup subscriptions with callback functions.

Now my callback functions are async in nature and thus cooroutine instead of normal functions because inside them i need to ```await``` other coroutines (in my case sending content to a websocket entpoint). Unfortunately the code inside ```pubsub.run()``` (which calls ```get_message()```  which calls ```handle_message()```) does not check if a callback might be an awaitable and should be awaited instead of called normally.

Instead the coroutine is never awaited, which raises a Warning and of course does not execute code inside the callback.

### To Reproduce

setup a pubsub with a callback subscription. The callback is a ```async def ...```
publish a message

### Expected behavior

The awaitable callback function should be awaited instead of directly called.

### Logs/tracebacks

```python-traceback
..../lib/python3.10/site-packages/aioredis/client.py:4177: RuntimeWarning: coroutine 'websocket_endpoint.<locals>.send_messages' was never awaited
  handler(message)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

### Python Version

```console
$ python --version
Python 3.10.0
```

### aioredis Version

```console
$ python -m pip show aioredis
Version 2.0.0
```

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Andrew-Chen-Wang:
I can't recall from top of my head, but does order matter when it comes to callbacks? If so, then all passed in callbacks must convert to coroutines which is breaking change. If not, then we can use `asyncio.iscoroutinefunction` to decide whether to await or not (I believe awaiting does slow down code by a little if no I/O is being done and comparing to a sync function that also does no I/O).

If that didn't make sense, I'm just asking you to file a PR and I'll assign proper reviewers. Thanks.

stephanm:
This didn't work for my use case, too.

@Andrew-Chen-Wang, maybe I didn't fully understand your concerns, but IMHO the order of the callbacks isn't guaranteed anyway as they are stored in a dict (unordered!) keyed by the channel. Maybe you could have a look and assign a reviewer?
Thanks!

PR #1284, opened 2022-02-02
Support for async callbacks

stephanm:
## What do these changes do?

As mentioned in #1235 you can't use coroutines as custom callbacks. This patch tries to fix that.

There shouldn't be any changes for the users as sync callbacks are still supported.
## Related issue number

#1235 
## Checklist

### DIFF
--- a/CHANGES/1284.feature
+++ b/CHANGES/1284.feature
@@ -1,0 +1,1 @@
+Using coroutines as callbacks is now possible.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -63,6 +63,7 @@ Sean Stewart <seandstewart>
 Sergey Miletskiy
 SeungHyun Hwang
 Stanislau Arkhipenka
+Stephan Meier
 Taku Fukada
 Taras Voinarovskyi
 Thanos Lefteris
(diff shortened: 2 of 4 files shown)


Andrew-Chen-Wang (changes requested):
Can you add a test where you receive two callbacks, one is sync and the other an async func? Thanks a ton! Also, pertaining to ordering, I can't actually remember my digging around of the code. In relation to your comment, dicts should be ordered now in supported Py3 versions, but I suppose it doesn't matter here 😅

stephanm:
Yeah, no problem, I'll add the test and have a look at the linter error.
Thanks for the clarifications and the fast feedback!
TIL: Ordering is guaranteed since 3.7 :)

stephanm:
I added a test case that uses both callback variants.

The typing problem wasn't that easy to fix.
Using something like
`PubSubHandler = Callable[[Dict[str, str]], Union[None, Awaitable[None]]]`
doesn't work:
`aioredis/client.py:4236: error: Incompatible types in "await" (actual type "Optional[Awaitable[None]]", expected type "Awaitable[Any]")  [misc]`
Maybe someone who is better at this typing stuff can comment on this issue? The current variant at least works with mypy :)

seandstewart (approved):
Thanks for your contribution here!

PR #1287, opened 2022-02-06
[WIP] Use asyncio protocol

seandstewart:
This supersedes #1216 and is an attempt to address the latency issues in #1208 -

Mainly I've gone through and added a Python parser which matches the hiredis implementation as closely as possible. I've also made a few adjustments to the RedisProtocol as introduced by @m-novikov in #1216.

I spent all day yesterday trying to get perfomance for this port to match something close to what @artesby & @m-novikov reported in their benchmarks in #1208, but I have a suspicion it's not possible.

@m-novikov and @artesby I'd be much obliged if you could test this implementation to see if we're actually going the right direction with this change. It's not yet ready for the real world and there are a few failing tests, but I don't want to keep barking up this tree if it won't get us where we need to go.

### DIFF
--- a/aioredis/client.py
+++ b/aioredis/client.py
@@ -1100,28 +1100,31 @@ class Redis:
         command_name = args[0]
         conn = self.connection or await pool.get_connection(command_name, **options)
         try:
-            await conn.send_command(*args)
-            return await self.parse_response(conn, command_name, **options)
+            resp = await conn.send_command(*args)
+            return await self.parse_response(resp, command_name, **options)
+        except ResponseError:
+            if EMPTY_RESPONSE in options:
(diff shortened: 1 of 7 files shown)


bmerry (commented):
Sorry, I'm now in a different team that's not working with aioredis, so I'm not available for reviewing.

Andrew-Chen-Wang:
thanks for all the help along the way @bmerry !

matemax:
Hi guys. Any update? Performance degradation is blocks an update to 2.* version in our projects =(

Andrew-Chen-Wang:
@matemax all updates are going to redis/redis-py. This repo at some point will convert back to its 1.3.1 version

PR #1313, opened 2022-03-04
Restore connection on network errors

evgenymarkov:
## Problem

If the application is isolated from the Internet, and then access to the Internet is returned, it will no longer recover due to the fact that it will use broken connections.

## What do these changes do?

Broken connections in which the StreamReader has an exception (e.g. `OSError No route to host`) are removed from the connection pool and are no longer reused.

Broken connections will be removed from connection pools.

### DIFF
--- a/CHANGES/1314.bugfix
+++ b/CHANGES/1314.bugfix
@@ -1,0 +1,1 @@
+Drop broken connections after network errors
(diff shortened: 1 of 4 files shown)


evgenymarkov:
I created issue https://github.com/aio-libs/aioredis-py/issues/1314

